PR #115, opened 2024-04-18
CHAIN-104 trigger block mining from tests

imykolenko:


integrationtests/src/test/kotlin/co/chainring/integrationtests/testutils/TestBlockchainClient.kt
```diff
@@ -19,17 +21,48 @@ class TestBlockchainClient(val config: BlockchainClientConfig = BlockchainClient
...
+        }
     }
 
     fun signData(hash: ByteArray): EvmSignature {
         val signature = Sign.signMessage(hash, credentials.ecKeyPair, false)
         return EvmSignature((signature.r + signature.s + signature.v).toHex())
     }
+
+    fun mine(numberOfBlocks: Int = config.numConfirmations) {
+        (1..numberOfBlocks).forEach { _ ->
```
bholzman, line 41: doesn't `anvil_mine` take an optional number of blocks to mine? seems like that would be better than calling it `numberOfBlocks` times

bholzman (approved):
Looks good!

integrationtests/src/test/kotlin/co/chainring/integrationtests/testutils/TestBlockchainClient.kt
imykolenko: good point, of course number of blocks can be passed. Changed.

integrationtests/src/test/kotlin/co/chainring/integrationtests/testutils/Wallet.kt
```diff
@@ -44,28 +44,40 @@ class Wallet(
     }
 
     fun mintERC20(symbol: String, amount: BigInteger) {
-        loadErc20Contract(symbol).mint(address.value, amount).send()
+        confirmedBlock {
```
antonyshchenko, line 47: Instead of wrapping here, I propose to add an extension function to `RemoteCall`, e.g:

```
loadErc20Contract(symbol).mint(address.value, amount).sendAndWaitForConfirmation()
```

imykolenko: changed

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -78,7 +78,7 @@ class BlockchainTransactionHandler(
                                         pendingTx.markAsSeen(blockNumber = blockNumber)
                                     }
                                 }
-                                if (numConfirmations >= confirmationsReceived) {
```
imykolenko, line 81: this line possibly caused occasional issues making tests highly sensitive to timing

PR #116, opened 2024-04-18
Broadcast order book updates over web-sockets

antonyshchenko:
Order book projection is reconstructed from the database by Broadcaster whenever it receives `GlobalNotification.OrderBook`. `ExchangeService` sends this event to `Broacaster` whenever it processes sequencer response and there were any created/updated orders.

### DIFF
--- a/backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
+++ b/backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
@@ -32,6 +32,7 @@ import org.jetbrains.exposed.sql.transactions.TransactionManager
 import org.jetbrains.exposed.sql.update
 import java.math.BigDecimal
 import java.math.BigInteger
+import java.math.RoundingMode
 
 @Serializable
 @JvmInline
@@ -258,16 +259,18 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
             }.firstOrNull()
         }
 
-        fun getOrdersMarkets(sequencerOrderIds: List<Long>): Set<MarketId> {
+        fun getOrdersMarkets(sequencerOrderIds: List<Long>): Set<MarketEntity> {
             return if (sequencerOrderIds.isEmpty()) {
                 emptySet()
             } else {
-                OrderTable
+                val marketIds = OrderTable
                     .select(OrderTable.marketGuid)
                     .where { OrderTable.sequencerOrderId.inList(sequencerOrderIds) }
                     .distinct()
                     .map { it[OrderTable.marketGuid].value }
                     .toSet()
+
+                MarketEntity.find { MarketTable.guid.inList(marketIds) }.toSet()
             }
         }
 
@@ -302,8 +305,11 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
             }
         }
 
+        fun getOrderBooks(markets: List<MarketEntity>): List<OrderBook> =
+            markets.map { getOrderBook(it) }
+
         fun getOrderBook(market: MarketEntity): OrderBook {
-            val priceScale = market.tickSize.stripTrailingZeros().scale()
+            val priceScale = market.tickSize.stripTrailingZeros().scale() + 1
 
             fun getOrderBookEntries(side: OrderSide): List<OrderBookEntry> {
                 val sizeCol = OrderTable.amount.sum().alias("size")
@@ -363,7 +369,7 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
                 buy = getOrderBookEntries(OrderSide.Buy),
                 sell = getOrderBookEntries(OrderSide.Sell),
                 last = LastTrade(
-                    price = lastTradePrice.setScale(priceScale).toString(),
+                    price = lastTradePrice.setScale(priceScale, RoundingMode.HALF_EVEN).toString(),
                     direction = when {
                         lastTradePrice >= prevTradePrice -> LastTradeDirection.Up
                         else -> LastTradeDirection.Down
--- a/backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
+++ b/backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
@@ -1,10 +1,15 @@
 package co.chainring.core.model.db
 
 import co.chainring.apps.api.model.Order
+import co.chainring.apps.api.model.websocket.LastTrade
+import co.chainring.apps.api.model.websocket.LastTradeDirection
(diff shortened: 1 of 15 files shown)


integrationtests/src/test/kotlin/co/chainring/integrationtests/api/OrderRoutesApiTest.kt
```diff
@@ -583,6 +679,29 @@ class OrderRoutesApiTest {
...
+                message.data.let { data ->
+                    assertIs<OrderBook>(data)
+                    assertEquals(
+                        OrderBook(
+                            marketId = btcEthMarketId,
+                            buy = listOf(
+                                OrderBookEntry(price = "17.50", size = "0.00012345".toBigDecimal()),
+                            ),
+                            sell = listOf(
+                                // TODO: amount should be reduced since MM order was partially filled
```
antonyshchenko, line 694: Discovered that we never update amounts of partially filled orders in DB

docker-compose.yaml
```diff
@@ -25,6 +25,16 @@ services:
       PGUSER: chainring # is used by psql inside container
     ports:
       - "5432:5432"
+  ut_db:
```
imykolenko, line 28: please, add same to `docker-compose-ci.yaml`, I created it recently to exclude otterscan image from being started on CI runner

backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
```diff
@@ -407,7 +420,13 @@ class ExchangeService(
 
         // queue any blockchain txs for processing
         blockchainClient.queueTransactions(blockchainTxs)
-        broadcasterNotifications.map { PrincipalNotifications(it.key, it.value) }.send()
+        NotificationsBatch(
+            principalNotifications = broadcasterNotifications.map { PrincipalNotifications(it.key, it.value) },
+            globalNotifications = OrderEntity
+                .getOrdersMarkets(response.ordersChangedList.map { it.guid })
+                .map { GlobalNotification.OrderBookUpdated(it) }
+                .sortedBy { it.marketId },
```
imykolenko, line 428: deposits and some withdrawals my result in no orders changed. In such case loading markets is unnecessary db call

imykolenko (approved):
Looks great

docker-compose.yaml
antonyshchenko: done

imykolenko:
Just as idea, sequencer already has all required information in memory including `totalQuantity` per level. This is maximum available data resolution. Querying same information from DB every second might be expensive. It would be great if we can just use already available in-memory state from sequencer in non-blocking (off-queue) way.
 
UI on the other hand in order to draw graph barely needs quantity per every single tick. It could potentially request specific `tick size view` (resolution) and zoom level like +- 10% from price. Having full resolution data in memory on API side would easy allow to reduce data to some partial view and in this way reduce amount of data flying around.
Looking far more in the future probably sending deltas would be good approach.

antonyshchenko:
@imykolenko we don't send order book to UI periodically anymore. I happens as a reaction on order changes while processing the sequencer response and the data is cached. 

I like your idea to get order book projection from sequencer itself without involving the database, though. We could include the new sizes for the affected levels into the sequencer response struct as well as last trade's price. Then when starting the backend process we could construct the order book projection from db once and update it incrementally with data from sequencer responses. @bholzman what do you think?

backend/src/main/kotlin/co/chainring/core/model/BroadcastNotifications.kt
```diff
@@ -38,11 +38,24 @@ sealed class BroadcasterNotification {
 
     @Serializable
     @SerialName("Orders")
-    data object Orders : BroadcasterNotification()
+    data object AllOpenOrdersCancelled : BroadcasterNotification()
```
bholzman, line 41: SerialName `Orders` seems like it should be changed

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
```diff
@@ -279,6 +301,76 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
                 it[this.closedAt] = now
             }
         }
+
+        fun getOrderBook(market: MarketEntity): OrderBook {
+            val priceScale = market.tickSize.stripTrailingZeros().scale()
```
bholzman, line 306: I think the price scale should be one greater than the tick size scale

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
```diff
@@ -279,6 +301,76 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
...
+                            size = size.toBigInteger().fromFundamentalUnits(market.baseSymbol.decimals).stripTrailingZeros(),
+                        )
+                    }
+            }
+
+            // We calculate last trade's price as a size-weighted average
+            // of all execution prices from the last match
+            val weightedAveragePriceCol = TradeTable.price.times(TradeTable.amount).sum().div(TradeTable.amount.sum())
+
+            val (lastTradePrice, prevTradePrice) = OrderExecutionTable
```
bholzman, line 341: seems like either we could keep the prev trade price in memory, or perhaps just determine the direction based on the side of the taker order -- direction must be up if it was a buy, and down if it was a sell

bholzman (approved):
Looks good!

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
antonyshchenko: Why so?

bholzman: Because the price can be between ticks. Especially if the market is tight and has bids and offers 1 tick apart, I think it can be useful to see if the price is closer to the bid or the offer.

antonyshchenko: That makes sense, thanks!

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
bholzman: Or maybe this is fine as is. Did realize, though, that we should add a LastDirection.Unchanged

antonyshchenko: Ok, leaving as is for now, will create a ticket for adding `LastDirection.Unchanged`

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
antonyshchenko: changed

PR #117, opened 2024-04-18
confirm deposit after number of confirmations

imykolenko:
Deposit is passed to sequencer after configurable number of confirmations


bflood366 (approved):
looks great

PR #118, opened 2024-04-18
CHAIN-119: write broadcast notification to DB and send guid as notify…

bflood366:
… payload


backend/src/main/kotlin/co/chainring/core/websocket/Broadcaster.kt
```diff
@@ -417,28 +341,22 @@ class Broadcaster(val db: Database) {
     }
 
     private fun sendBalances(principal: Principal) {
-        findClients(principal, SubscriptionTopic.Orders)
```
imykolenko, line 420: good catch

backend/src/main/kotlin/co/chainring/core/websocket/Broadcaster.kt
```diff
@@ -417,28 +341,22 @@ class Broadcaster(val db: Database) {
...
 
     private fun handleDbNotification(payload: String) {
         logger.debug { "received db notification with payload $payload" }
         try {
-            val notifications = Json.decodeFromString<List<PrincipalNotifications>>(payload)
-            notifications.forEach { principalNotifications ->
+            val notificationData = transaction {
+                BroadcasterNotificationEntity.findById(BroadcasterNotificationId(payload))?.notificationData
+            }
+            notificationData?.forEach { principalNotifications ->
```
imykolenko, line 354: I know sending login was done outside if this PR, but I propose to make few tweaks while this code changes:
 - process principals in parallel, but send in sequence per principal. This should decrease time to message sent
  - catch exceptions per principal, just in case any particular socket is closed rest of the batch will be sent out

imykolenko (approved):
Looks great to me!

backend/src/main/kotlin/co/chainring/core/model/BroadcasterNotification.kt
```diff
@@ -0,0 +1,23 @@
...
+@Serializable
+sealed class BroadcasterNotification {
+
+    @Serializable
+    @SerialName("Balances")
+    data object Balances : BroadcasterNotification()
+
+    @Serializable
+    @SerialName("Orders")
+    data object Orders : BroadcasterNotification()
```
antonyshchenko, line 16: Why does these two have to be separate classes? Why not just use `Publishable` for any kind of a notification?

backend/src/main/kotlin/co/chainring/core/model/db/BroadcasterNotification.kt
```diff
@@ -0,0 +1,50 @@
...
+
+@JvmInline
+value class BroadcasterNotificationId(override val value: String) : EntityId {
+    companion object {
+        fun generate(): BroadcasterNotificationId = BroadcasterNotificationId(TypeId.generate("bcnotif").toString())
+    }
+    override fun toString(): String = value
+}
+
+object BroadcasterNotificationTable : GUIDTable<BroadcasterNotificationId>("broadcaster_notification", ::BroadcasterNotificationId) {
```
antonyshchenko, line 28: I think this is more of a job, than a notification. How about renaming the table and entity to `BroadcasterJob`? Otherwise it is a little confusing that `broadcaster_notification` table contains a list of notifications per receiver. We are using the term notification in 2 different contexts here and this leads to ambiguity and confusion.

It might also make sense to somehow mark these as processed or alternatively we can delete the record after it has been handled by the `Broadcaster`. Deleting might be a better option as we would avoid accumulating garbage in the database.

Also I'm not sure what should we do with records that were created while the API was down. Broadcaster has missed the `NOTIFY` events in this case, but the records would be present in this table. We certainly should not try to deliver those messages since websockets were disconnected anyway, but I don't think we should keep them in the database either.

backend/src/main/kotlin/co/chainring/core/websocket/Broadcaster.kt
bflood366: I made the change to catch exceptions per principal. Will add a separate ticket for sending in parallel per principal

backend/src/main/kotlin/co/chainring/core/model/db/BroadcasterNotification.kt
bflood366: ok I change this to BroadcasterJob. For now it not being deleted. Will add a separate ticket for that.

PR #119, opened 2024-04-19
CHAIN-116: fix broken css in prod build

bholzman:


imykolenko (approved):
👍 

Right, so dynamically generated classes were purged. Great find.

PR #124, opened 2024-04-22
Remove unnecessary responsibilities from blockchain client

antonyshchenko:
I think `BlockchainClient` should not contain any db logic or threads and should rather be a thin client

- `BlockchainTransactionHandler` and `BlockchainDepositHandler` are separate components that use BlockchainClient
- Removed circular dependency between `BlockchainClient` and `BlockchainTransactionHandler`/`BlockchainDepositHandler`
- Extracted `ContractsPublisher` from `BlockchainClient`


bflood366 (approved):
looks great

PR #131, opened 2024-04-22
Reduce BlockchainClient noise in logs

antonyshchenko:
- when refreshing transactions in `BlockchainTransactionHandler` request block number only if there are any pending transactions
- when refreshing deposits in `BlockchainDepositHandler` request block number only if there any pending deposits
- log RPC calls in `BlockchainClient` with one log statement
- filter out RPC calls to get exchange contract events unless we actually get any events in response (`BlockchainDepositHandler` makes this call every 300 ms)


bholzman (approved):
Looks good!

PR #132, opened 2024-04-23
Update order amount in DB when partially filled

antonyshchenko:
Changes in `BlockchainTransactionHandler` and `BlockchainDepositHandler` are caused by sporadic test failures due to db cleanup before each test while the app is running in background.


bholzman (approved):
Looks great!

PR #136, opened 2024-04-23
Bflood/chain 97 - process updates from the chronicle queue

bflood366:
- `ExchangeService` has been split into 2 - `ExchangeService` handles all calls from API server , `QueueProcessorService` is called from `QueueProcessorApp` which runs in sequencer process and handles all logic based on consuming the chronicle response/request queues
- All order related APIs are asynchronous - they will returns as soon as sequencer has accepted request. For batch operations there  is now a `requestStatus` as part of each batch entity in the response to indicate if the specific create/update/cancel was accepted or rejected by the sequencer.  For single entity API operations (createOrder, updateOrder or cancelOrder- the api call will fail with a 4xx code if the sequencer rejects the single operation)
- I ran into issue where both checksummed and non checksummed address were being sent from web-ui which caused us to treat that address as 2 different wallets. Now all addresses are stored as checksummed
- I have added logic to the sequencer to validate the order belongs to the requesting wallet before updating or cancelling.


backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
```diff
@@ -94,207 +83,165 @@ class ExchangeService(
...
-
+    ): BatchOrdersApiResponse {
         if (apiRequest.cancelOrders.isEmpty() && apiRequest.updateOrders.isEmpty() && apiRequest.createOrders.isEmpty()) {
-            return emptyList()
+            return BatchOrdersApiResponse(emptyList(), emptyList(), emptyList())
         }
 
-        val sequencerOrderIdsBeingUpdated = mutableListOf<SequencerOrderId>()
-        val updatedAndCancelsOrderIds = mutableListOf<OrderId>()
+        val createdOrders = mutableMapOf<OrderId, CreateOrderApiRequest>()
```
antonyshchenko, line 112: No need for `mutableMap` here. Could be replaced with:
```
val createOrderRequestsByOrderId = apiRequest.createOrders.associateBy { OrderId.generate() }
val ordersToAdd = createOrderRequestsByOrderId.map { (orderId, request) ->
   ...
   SequencerClient.Order(...)
}
```

backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
```diff
@@ -606,20 +444,13 @@ class ExchangeService(
 
     override fun onExchangeContractDepositConfirmation(deposit: DepositEntity) {
         transaction {
-            val response = runBlocking {
-                sequencerClient.deposit(deposit.wallet.sequencerId.value, Asset(deposit.symbol.name), deposit.amount)
+            runBlocking {
```
antonyshchenko, line 447: `transaction {` could be removed since `BlockchainDepositHandler` calls this in the transaction context

antonyshchenko: Also seems like `queueBlockchainExchangeTransactions` method could be removed from this class

backend/src/main/kotlin/co/chainring/core/services/QueueProcessorService.kt
```diff
@@ -0,0 +1,305 @@
...
+import co.chainring.sequencer.proto.SequencerError
+import co.chainring.sequencer.proto.SequencerRequest
+import co.chainring.sequencer.proto.SequencerResponse
+import co.chainring.sequencer.proto.newQuantityOrNull
+import io.github.oshai.kotlinlogging.KotlinLogging
+import kotlinx.datetime.Clock
+import org.jetbrains.exposed.sql.transactions.transaction
+import java.math.BigInteger
+
+object QueueProcessorService {
```
antonyshchenko, line 56: I would argue that this should be moved into `co.chainring.sequencer.apps` and I would rename it to `SequencerResponseProcessor` or something like that, since queue processor is a little too generic.

backend/src/main/kotlin/co/chainring/core/services/QueueProcessorService.kt
```diff
@@ -0,0 +1,305 @@
...
+
+    fun updateLastProcessedIndex(lastProcessedIndex: Long) = transaction {
+        QueueProcessorEntity.findQueueProcessor()?.let {
+            it.update(lastProcessedIndex)
+        } ?: run {
+            QueueProcessorEntity.create(lastProcessedIndex)
+        }
+    }
+
+    fun onSequencerResponseReceived(response: SequencerResponse, request: SequencerRequest) {
```
antonyshchenko, line 76: Is there a reason not to wrap the entire `onSequencerResponseReceived` with `transaction`? How do we recover if processing one of the order responses in the batch fails?

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
```diff
@@ -0,0 +1,90 @@
...
+import co.chainring.sequencer.proto.SequencerRequest
+import co.chainring.sequencer.proto.SequencerResponse
+import io.github.oshai.kotlinlogging.KotlinLogging
+import net.openhft.chronicle.queue.impl.RollingChronicleQueue
+import java.lang.Thread.UncaughtExceptionHandler
+import kotlin.concurrent.thread
+
+val dbConfig: DbConfig = DbConfig()
+
+class QueueProcessorApp(
```
antonyshchenko, line 15: The name is slightly misleading since it processes the output queue.

sequencer/src/main/kotlin/co/chainring/sequencer/Main.kt
```diff
@@ -21,6 +22,11 @@ fun main(args: Array<String>) {
             },
         )
         sequencer.start()
+        val queueProcessorApp = QueueProcessorApp(
```
antonyshchenko, line 25: Could be a separate ticket, but I think we should move `SequencerApp` into a separate process/container so that it does not get affected by GC pauses in `GatewayApp` and `QueueProcessorApp`.

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
```diff
@@ -0,0 +1,90 @@
...
+                            dc.wire()?.read()?.bytes { bytes ->
+                                try {
+                                    val request = SequencerRequest.parseFrom(bytes.toByteArray())
+                                    logger.debug { "processing sequence ${resp.sequence} request = <$request> response=<$resp" }
+                                    QueueProcessorService.onSequencerResponseReceived(
+                                        resp,
+                                        SequencerRequest.parseFrom(bytes.toByteArray()),
+                                    )
+                                } catch (e: Exception) {
+                                    logger.error(e) { "unhandled exception processing ${resp.sequence}" }
```
antonyshchenko, line 62: Do we really want to just continue here? Most probably processing subsequent responses will result in more failures and inconsistent db state.

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
```diff
@@ -0,0 +1,90 @@
...
+                                } finally {
+                                    logger.debug { "storing last processed index ${outputTailer.lastReadIndex()}" }
+                                    QueueProcessorService.updateLastProcessedIndex(outputTailer.lastReadIndex())
```
antonyshchenko, line 65: We might want to do this in the same database transaction where we process the response.

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainDepositHandler.kt
```diff
@@ -100,6 +101,7 @@ class BlockchainDepositHandler(
                                     amount = amount,
                                     blockNumber = blockNumber,
                                     transactionHash = txHash,
+                                    randomLong(),
```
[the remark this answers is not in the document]
bflood366: actually removed this - was not needed

backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
```diff
@@ -85,7 +74,7 @@ class ExchangeService(
     fun addOrder(
         walletAddress: Address,
         apiRequest: CreateOrderApiRequest,
-    ): Order {
+    ): CreateOrderApiResponse {
```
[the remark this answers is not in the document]
bflood366: actually it was also used for confirmation callbacks so I split it into an `ExchangeApiService` which I moved into the `co.chainring.apps.api.services` package and an `ExchangeConfirmationService` which I left in the `co.chainring.core.services` package

backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
bflood366: fixed, thx

backend/src/main/kotlin/co/chainring/core/services/QueueProcessorService.kt
bflood366: moved and renamed

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
bflood366: renamed to `SequencerResponseProcessorApp`

sequencer/src/main/kotlin/co/chainring/sequencer/Main.kt
bholzman: That's a good idea - created https://chainringinc.atlassian.net/browse/CHAIN-141 for this

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
bflood366: made the change to do a specific number of retries and then stop on if retry count exceeded

sequencer/src/main/kotlin/co/chainring/sequencer/apps/QueueProcessorApp.kt
bflood366: changed

backend/src/main/kotlin/co/chainring/core/services/QueueProcessorService.kt
bflood366: everything on one transaction now

backend/src/main/kotlin/co/chainring/core/services/ExchangeService.kt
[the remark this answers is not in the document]
bflood366: changed this per your suggestion

PR #137, opened 2024-04-24
use duration enum values instead of ms

imykolenko:
- duration enum instead of ms for OHLC messages
- fix ohlc gap filling on other intervals than 5m
- re-fetch full prices when pg listener reconnects

### DIFF
--- a/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
@@ -16,7 +16,7 @@ import {
   pricesTopic,
   Publishable
 } from 'websocketMessages'
-import { mergeOHLC } from 'utils/pricesUtils'
+import { mergeOHLC, olhcDurationsMs } from 'utils/pricesUtils'
 import { useWebsocketSubscription } from 'contexts/websocket'
 import { useWindowDimensions, widgetSize, WindowDimensions } from 'utils/layout'
 import { produce } from 'immer'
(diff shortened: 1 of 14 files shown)


web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
```diff
@@ -27,22 +40,13 @@ type PriceParameters = {
...
-  [OHLCDuration.P1H]: 60 * 60 * 1000,
-  [OHLCDuration.P4H]: 4 * 60 * 60 * 1000,
-  [OHLCDuration.P1D]: 24 * 60 * 60 * 1000
+const durationsMs: Record<OHLCDuration, number> = {
+  ['P1M']: 60 * 1000,
+  ['P5M']: 5 * 60 * 1000,
+  ['P15M']: 15 * 60 * 1000,
+  ['P1H']: 60 * 60 * 1000,
+  ['P4H']: 4 * 60 * 60 * 1000,
+  ['P1D']: 24 * 60 * 60 * 1000
```
antonyshchenko, line 49: Seems like this could be moved to where `mergeOHLC` is defined and `mergeOHLC` could expect `OHLCDuration` as argument instead of number representing duration in milliseconds.

antonyshchenko (approved):
Looks great!

web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
imykolenko: makes sense, updated

PR #142, opened 2024-04-25
CHAIN-118 calculate weighted price to update ohlc

imykolenko:


sequencer/src/main/kotlin/co/chainring/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -261,16 +262,23 @@ object SequencerResponseProcessorService {
...
-            OHLCEntity.updateWith(trade).map {
-                BroadcasterNotification.pricesForMarketPeriods(
-                    trade.marketGuid.value,
-                    it.duration,
-                    listOf(it),
-                    full = false,
-                )
-            }
-        }.flatten().distinct()
+        val ohlcNotifications = tradeEntities.groupBy { it.market }
```
bholzman, line 265: I didn't mean to use weighted price over all trades in a given market within the sequencer response, rather I meant for all executions resulting from a distinct market order. But I guess mostly it will be the same since it is probably unlikely that an order batch would have more than one market order in the same market. So probably fine to leave it this way for now.

bholzman (approved):
Looks good

PR #144, opened 2024-04-25
CHAIN-142 prices chart auto pan

imykolenko:
- make price chart follow new candles in case when panned right 
 - reduce frequency of labels to fit more candlesticks (atm max 30, can also be dynamic) 
 
========
Screenshot 2024-04-25 at 1 46 04 PM
Screenshot 2024-04-25 at 1 52 31 PM

### DIFF
--- a/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
@@ -16,7 +16,7 @@ import {
   pricesTopic,
   Publishable
 } from 'websocketMessages'
-import { mergeOHLC, olhcDurationsMs } from 'utils/pricesUtils'
+import { mergeOHLC, ohlcDurationsMs } from 'utils/pricesUtils'
 import { useWebsocketSubscription } from 'contexts/websocket'
 import { useWindowDimensions, widgetSize, WindowDimensions } from 'utils/layout'
 import { produce } from 'immer'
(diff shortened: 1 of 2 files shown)


web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
```diff
@@ -292,13 +289,23 @@ export function PricesWidget({ marketId }: { marketId: string }) {
...
-            date.getHours() + ':' + (minutes < 10 ? '0' + minutes : minutes)
-          )
+
+          const labelIntervalMinutes = ohlcDuration === 'P5M' ? 15 : 5
+          if (minutes % labelIntervalMinutes == 0) {
+            return (
+              date.getHours() + ':' + (minutes < 10 ? '0' + minutes : minutes)
+            )
+          } else {
+            return ''
```
bholzman, line 307: seems like it might be better to have label density be a function of the width of the widget and the number of bars, instead of being hard-coded based on the specific duration.

bholzman (approved):

web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
imykolenko: Agreed, this is a good point. By now I have another 200 lines of code that doing dynamic labeling ensuring that significant time points (hour / date / month change) are labeled. I'll simplify it a bit later and send in another PR. 

Merging.

PR #146, opened 2024-04-25
CHAIN-144: fixes to order book widget

bholzman:


mocker/src/main/kotlin/co/chainring/Main.kt
```diff
@@ -26,15 +26,15 @@ fun main() {
...
+                "USDC" to 100.toFundamentalUnits(6),
+                "DAI" to 50.toFundamentalUnits(18)
             )
         )
     }
     takers.forEach {
         it.start(listOf(usdcDai))
     }
-    Thread.sleep(60000)
+    Thread.sleep(600000)
```
bflood366, line 37: maybe default it to 600000 and also allow it to be passed in as a command line arg this way if we want to run this for a longer (or shorter) period we don't need to change the code.

bflood366 (approved):
looks great

mocker/src/main/kotlin/co/chainring/Main.kt
bholzman: I find myself making frequent edits here to try different scenarios, not sure command-line args would be the best way to parameterize -- thinking we need more like a yaml file to config. Will add a ticket but leave this as is.

antonyshchenko: I would rather have an option to run this indefinitely until I kill the process myself.

PR #147, opened 2024-04-25
CHAIN-146: don't allow partial market order to be changed or cancelle…

bflood366:
…d and fix trade amount display


antonyshchenko (approved):
Looks great!

PR #150, opened 2024-04-26
CHAIN-123: require signature on order update

bflood366:


backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
```diff
@@ -311,15 +324,21 @@ class ExchangeApiService(
...
-                    EIP712Helper.computeHash(
-                        tx,
-                        blockchainClient.chainId,
-                        verifyingContract,
-                    ),
-                    tx.signature,
-                    walletAddress,
-                )
+            if (
+                try {
```
antonyshchenko, line 328: Not sure how I feel about try/catch inside an if-statement. And in general I think this method has too much nesting which makes it hard to follow. Why don't we check the presence of exchange contract first and then wrap the rest into try/catch?
E.g something like this:

```
private fun validateOrderSignature(walletAddress: Address, marketId: MarketId, amount: BigInteger, side: OrderSide, price: BigDecimal?, nonce: String, signature: EvmSignature) {
  val verifyingContract = blockchainClient.getContractAddress(ContractType.Exchange) ?: throw ExchangeError("No deployed contract found")
  val isValidSignature = try {
    val (baseSymbol, quoteSymbol) = marketId.value.split("/")
    ...
    val hash = EIP712Helper.computeHash(...)
    ECHelper.isValidSignature(hash, tx.signature, walletAddress))
  } catch (e: Exception) {
    logger.warn(e) { "Exception validating signature" }
    throw ExchangeError("Invalid signature")
  }
  if (!isVaildSignature) {
    throw ExchangeError("Invalid signature")
  }
}
```

Or like this:
```
private fun validateOrderSignature(walletAddress: Address, marketId: MarketId, amount: BigInteger, side: OrderSide, price: BigDecimal?, nonce: String, signature: EvmSignature) {
  val verifyingContract = blockchainClient.getContractAddress(ContractType.Exchange) ?: throw ExchangeError("No deployed contract found")
  runCatching {
    val (baseSymbol, quoteSymbol) = marketId.value.split("/")
    ...
    val hash = EIP712Helper.computeHash(...)
    ECHelper.isValidSignature(hash, tx.signature, walletAddress))
  }.onFailure { e ->
    logger.warn(e) { "Exception validating signature" }
  }.getOrDefault(false).also { isValidSignature ->
    if (!isValidSignature) {
      throw ExchangeError("Invalid signature")
    }
  }
}
```

integrationtests/src/main/kotlin/co/chainring/integrationtests/utils/Wallet.kt
```diff
@@ -96,6 +97,17 @@ class Wallet(
         return request.copy(signature = blockchainClient.signData(EIP712Helper.computeHash(tx, blockchainClient.chainId, exchangeContractAddress)))
     }
 
+    fun signOrder(request: UpdateOrderApiRequest.Limit): UpdateOrderApiRequest.Limit {
+        val (baseSymbol, quoteSymbol) = request.marketId.value.split("/")
+        val tx = request.toEip712Transaction(
```
antonyshchenko, line 102: Seems like this could be simpler:

In `UpdateOrderApiRequest.Limit`:

```
fun toEip712Transaction(sender: WalletEntity, baseToken: SymbolEntity, quoteToken: SymbolEntity)
```

Then here:

```
va tx = request.toEip712Transaction(this, symbols.first { it.name == baseSymbol }, symbols.first { it.name == quoteSymbol })
```

web-ui/src/components/Screens/HomeScreen/OrdersAndTradesWidget.tsx
```diff
@@ -266,26 +275,45 @@ export default function OrdersAndTradesWidget({
...
   const quoteSymbol = market.quoteSymbol
+  const { signTypedDataAsync } = useSignTypedData()
 
-  const [amount, setAmount] = useState(
-    formatUnits(order.amount, baseSymbol.decimals)
-  )
-  const [price, setPrice] = useState(
-    order.type == 'limit' ? String(order.price) : ''
-  )
+  const {
```
antonyshchenko, line 298: 👍

antonyshchenko (approved):
Looks great!

integrationtests/src/main/kotlin/co/chainring/integrationtests/utils/Wallet.kt
bflood366: simplified per your suggestion

PR #154, opened 2024-04-30
CHAIN-137: require signature for cancelling orders

bflood366:
currently what must signed is sender address, MarketId of the order, remaining amount of the order and a nonce.

### DIFF
--- a/backend/src/main/kotlin/co/chainring/apps/api/model/Orders.kt
+++ b/backend/src/main/kotlin/co/chainring/apps/api/model/Orders.kt
@@ -38,7 +38,7 @@ sealed class CreateOrderApiRequest {
         override val amount: BigIntegerJson,
         override val signature: EvmSignature,
     ) : CreateOrderApiRequest() {
-        fun toEip712Transaction(sender: Address, baseToken: co.chainring.apps.api.model.Symbol, quoteToken: co.chainring.apps.api.model.Symbol) = EIP712Transaction.Order(
+        override fun toEip712Transaction(sender: Address, baseToken: co.chainring.apps.api.model.Symbol, quoteToken: co.chainring.apps.api.model.Symbol) = EIP712Transaction.Order(
             sender,
             baseToken.contractAddress ?: Address.zero,
             quoteToken.contractAddress ?: Address.zero,
@@ -59,7 +59,7 @@ sealed class CreateOrderApiRequest {
         val price: BigDecimalJson,
         override val signature: EvmSignature,
     ) : CreateOrderApiRequest() {
-        fun toEip712Transaction(sender: Address, baseToken: co.chainring.apps.api.model.Symbol, quoteToken: co.chainring.apps.api.model.Symbol) = EIP712Transaction.Order(
+        override fun toEip712Transaction(sender: Address, baseToken: co.chainring.apps.api.model.Symbol, quoteToken: co.chainring.apps.api.model.Symbol) = EIP712Transaction.Order(
             sender,
             baseToken.contractAddress ?: Address.zero,
             quoteToken.contractAddress ?: Address.zero,
@@ -75,6 +75,8 @@ sealed class CreateOrderApiRequest {
             is Market -> null
         }
     }
+
+    abstract fun toEip712Transaction(sender: Address, baseToken: co.chainring.apps.api.model.Symbol, quoteToken: co.chainring.apps.api.model.Symbol): EIP712Transaction.Order
 }
 
 @Serializable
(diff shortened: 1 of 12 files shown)


backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
```diff
@@ -225,18 +240,19 @@ class ExchangeApiService(
...
-                    createOrders = emptyList(),
-                    updateOrders = emptyList(),
-                    cancelOrders = listOf(CancelOrderApiRequest(orderId)),
-                ),
-            ).canceledOrders.first()
-        } ?: throw ExchangeError("Requested order does not exist")
+    fun cancelOrder(walletAddress: Address, orderId: OrderId, cancelOrderApiRequest: CancelOrderApiRequest): CancelOrderApiResponse {
+        if (orderId != cancelOrderApiRequest.orderId) {
+            throw ExchangeError("order Ids must match")
+        }
```
antonyshchenko, line 246: IMO, this check should be done in the `OrderRoutes` and `cancelOrder` method here should only take `cancelOrderApiRequest` as param, but not  both `orderId` and `cancelOrderApiRequest`

backend/src/main/kotlin/co/chainring/apps/api/model/Orders.kt
```diff
@@ -141,7 +141,20 @@ enum class RequestStatus {
 @Serializable
 data class CancelOrderApiRequest(
     val orderId: OrderId,
-)
+    val marketId: MarketId,
+    val side: OrderSide,
+    val amount: BigIntegerJson,
```
antonyshchenko, line 146: Why do we require client to provide these fields? Can't we look them up from the database in `ExchangeApiService`?

backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
```diff
@@ -318,21 +334,31 @@ class ExchangeApiService(
         }
     }
 
-    private fun validateOrderSignature(walletAddress: Address, marketId: MarketId, amount: BigInteger, side: OrderSide, price: BigDecimal?, nonce: String, signature: EvmSignature) {
+    private fun validateOrderSignature(walletAddress: Address, marketId: MarketId, amount: BigInteger, side: OrderSide, price: BigDecimal?, nonce: String, signature: EvmSignature, isCancel: Boolean = false) {
```
antonyshchenko, line 337: Not sure how I feel about `isCancel` flag. We already have `toEip712Transaction` defined on API request models. How about changing the signature of this method to expect `walletAddress`, `EIP712Transaction` and `EvmSignature` and then move the responsibility of transforming API request to `EIP712Transaction` to the caller?

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
```diff
@@ -123,6 +126,43 @@ sealed class EIP712Transaction {
...
+            return EIP712TransactionType.CancelOrder
+        }
+
+        override fun getModel(): List<StructuredData.Entry> {
+            val model = mutableListOf<StructuredData.Entry>()
+            model.add(StructuredData.Entry("sender", "address"))
+            model.add(StructuredData.Entry("marketId", "string"))
+            model.add(StructuredData.Entry("amount", "int256"))
+            model.add(StructuredData.Entry("nonce", "int256"))
+            return model
```
antonyshchenko, line 149: Why not `return listOf(StructuredData.Entry("sender", "address"), ..)`? This could be simplified even further into a `val model = listOf(StructuredData.Entry("sender", "address"), ..)`. Same applies for `getTransactionType()`, `getMessage()` and `getTxData()` - they don't have t be defined as methods, could be just abstract vals instead.

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
bflood366: changed

backend/src/main/kotlin/co/chainring/apps/api/model/Orders.kt
bflood366: We treat the order paths as a hot path (create, update or cancel) so we are not going to the DB in these paths - we verify the signature before we send to the sequencer, so going to the DB would slow things down. Also the client needs these fields for the signing so they have them anyway and can pass them in.

backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
bflood366: yep, great suggestion - changed

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
bflood366: changed to return a list per your suggestion, but left the methods for now - the EIP712Transaction is Serializable so if we wanted to make these vars we would need to make those serializable and `StructuredData.Entry` is not serializable.

PR #155, opened 2024-04-30
SSH Key loadtest chainring

imykolenko:
- ec2 instance for loadtest, type can be adjusted later
 - ssh key is in the 1Password`SSH Key loadtest chainring`
 SSH connection works via bastion (`ssh -J ec2-user@test-bastion.chainring.co 10.10.3.134`) by adding host to the ssh hosts
 - java is installed manually, as well as chainring project copied into home

EC2 instance still can't connect to `test-api.chainring.co`


antonyshchenko (approved):
Nice! Latency values are much more readable now!

PR #156, opened 2024-04-30
CHAIN-118 price labels on significant points of time

imykolenko:
- use cron expressions to place labels deterministically on specific minutes, hours
 - label significant points of time
 - increase number of candlesticks

=======
Screenshot 2024-04-30 at 9 44 42 PM
Screenshot 2024-04-30 at 9 44 24 PM
Screenshot 2024-04-30 at 9 48 18 PM
Screenshot 2024-04-30 at 9 48 40 PM

### DIFF
--- a/web-ui/src/utils/dateUtils.ts
+++ b/web-ui/src/utils/dateUtils.ts
@@ -7,7 +7,7 @@ import {
   SixtyRange
 } from 'cron-parser'
 
-export const isDateMatchesCronExpression = (
+export const doesDateMatchCronExpression = (
   expression: string,
   date: Date,
   scope: string = 'second'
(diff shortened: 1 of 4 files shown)


web-ui/src/utils/dateUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+      return false
+    if (
+      scopeIndex <= 5 &&
+      !data.dayOfWeek.includes(date.getUTCDay() as DayOfTheWeekRange)
+    )
+      return false
+
+    return true
+  } catch (e) {
+    throw new Error(`isDateMatchesCronExpression error: ${e}`)
```
antonyshchenko, line 55: Why do we have to catch and re-throw here?

web-ui/src/utils/dateUtils.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  DayOfTheMonthRange,
+  DayOfTheWeekRange,
+  HourRange,
+  MonthRange,
+  parseExpression,
+  SixtyRange
+} from 'cron-parser'
+
+export const isDateMatchesCronExpression = (
```
antonyshchenko, line 10: Nit: I would drop the word `is` from the method name to fix the grammar. Or alternatively this could be renamed to `doesDataMatchCronExpression`

antonyshchenko (approved):
Looks great!

web-ui/src/utils/dateUtils.ts
imykolenko: good point. Simplified, just returning false, should never happen anyway.

web-ui/src/utils/dateUtils.ts
imykolenko: renamed

PR #157, opened 2024-05-03
use default price if ohlc is empty

imykolenko:
In case when OHLC is empty for the market use default price to start the loadtest

### DIFF
--- a/mocker/src/main/kotlin/co/chainring/mocker/core/Taker.kt
+++ b/mocker/src/main/kotlin/co/chainring/mocker/core/Taker.kt
@@ -53,6 +53,7 @@ class Taker(
     private var stopping = false
 
     fun start(marketIds: List<MarketId>) {
+        marketIds.forEach { marketPrices[it] = 2.toBigDecimal() }
         val config = apiClient.getConfiguration()
         markets = config.markets.toSet()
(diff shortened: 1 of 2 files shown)


mocker/src/main/kotlin/co/chainring/mocker/core/Taker.kt
```diff
@@ -53,6 +53,7 @@ class Taker(
     private var stopping = false
 
     fun start(marketIds: List<MarketId>) {
+        marketIds.forEach { marketPrices[it] = 2.toBigDecimal() }
```
bholzman, line 56: We already require specifying an initial price when creating a market in the sequencer -- maybe it would be better to store that in the DB and use that here instead of a hard-coded 2?

bholzman (approved):

mocker/src/main/kotlin/co/chainring/mocker/core/Taker.kt
imykolenko: Yeah, you right, hardcoding price it is really a quick fix. Hm.. with crossing orders initial price can be arbitrary, theoretically. Another option is to seed OHLC, currently records are deleted when cleaning DB otherwise testing is time sensitive.

Clothing this PR for now.

PR #161, opened 2024-05-07
Automatically add network to wallet

antonyshchenko:
https://github.com/Chainring-Inc/chainring-monorepo/assets/443906/6fb63033-29f4-4c5f-aa98-260328511448


web-ui/src/components/Screens/Header.tsx
```diff
@@ -23,13 +24,23 @@ export function Header({
...
   useEffect(() => {
     if (account.isConnected && account.connector) {
       setIcon(account.connector.icon)
       setName(account.connector.name)
+
+      switchChain({
+        addEthereumChainParameter: {
+          chainName: chain.name,
+          nativeCurrency: chain.nativeCurrency,
+          rpcUrls: chain.rpcUrls.default.http
```
bholzman, line 38: should we add an optional blockExplorerUrl to our chain config and set it here if we have it available?

bholzman (approved):
Looks good!

web-ui/src/components/Screens/Header.tsx
antonyshchenko: good point, added

PR #162, opened 2024-05-07
Feature/chain 145 allow limit order to cross the market

imykolenko:


sequencer/src/main/kotlin/co/chainring/sequencer/core/OrderBook.kt
```diff
@@ -227,6 +227,7 @@ data class RemoveOrderResult(
 data class ChangeOrderResult(
     val wallet: WalletAddress,
     val disposition: OrderDisposition,
+    val executions: List<Execution>,
     val baseAssetDelta: BigInteger,
     val quoteAssetDelta: BigInteger,
 )
```
imykolenko, line 233: change order now might result into immediate executions

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
```diff
@@ -143,65 +165,38 @@ data class Market(
...
-                val quote = id.quoteAsset()
-                val (buyer, seller) = if (order.type == Order.Type.MarketBuy) wallet to execution.counterOrder.wallet else execution.counterOrder.wallet to wallet
-                balanceChanges.merge(Pair(buyer, quote), -notional, ::sumBigIntegers)
-                balanceChanges.merge(Pair(seller, base), -execution.amount, ::sumBigIntegers)
-                balanceChanges.merge(Pair(buyer, base), execution.amount, ::sumBigIntegers)
-                balanceChanges.merge(Pair(seller, quote), notional, ::sumBigIntegers)
-                if (order.type == Order.Type.MarketBuy) {
-                    consumptionChanges.merge(execution.counterOrder.wallet, Pair(-execution.amount, notional), ::sumBigIntegerPair)
-                } else {
-                    consumptionChanges.merge(execution.counterOrder.wallet, Pair(execution.amount, -notional), ::sumBigIntegerPair)
```
imykolenko, line 203: When processing executions it looks like the reserved consumption should only decrease.
I updated to 
```
buy -> (-amount, 0)
sell -> (0, -notional)
```

bholzman: ah, good catch ... the balance on the other side changes but not the reserved

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
```diff
@@ -286,26 +329,36 @@ data class Market(
         notional(it.quantity, levels[it.levelIx].price, baseDecimals, quoteDecimals)
     }?.reduceOrNull(::sumBigIntegers) ?: BigInteger.ZERO
 
-    private fun handleMarketOrder(order: Order): AddOrderResult {
+    private fun handleMarketOrder(order: Order, stopAtLevelIx: Int? = null): AddOrderResult {
```
bholzman, line 332: perhaps rename to `handleCrossingOrder` ?

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
```diff
@@ -380,51 +433,75 @@ data class Market(
...
+                    if (remainingAmount > BigInteger.ZERO) {
+                        // and then create limit order for the remaining amount
+                        val adjustedOrder = order.copy { amount = remainingAmount.toIntegerValue() }
+                        val limitOrderDisposition = createLimitSellOrder(levelIx, wallet, adjustedOrder)
+
+                        AddOrderResult(
+                            // market part of the limit order can be rejected due to no liquidity yet in the empty market
+                            // this happens because `bestBid` set to half tick away from the market price during market creation
+                            // potentially can be solved by widening initial bestBid/bestOffer distance (e.g. outer boundaries of the market)
+                            if (marketOrderResult.disposition == OrderDisposition.Rejected) limitOrderDisposition else marketOrderResult.disposition,
```
bholzman, line 459: hmm, if there's no liquidity then there's no market part of the limit order - should just become a limit order on the book (and the bestBid/bestOffer should be changed) - I wonder if we can adjust logic so that bestBid and bestOffer are null until there's actually an order on that side of the book

bholzman (approved):
Looks great!

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
imykolenko: Order sent to a new empty it does become a limit order, it is just requires this  additional step to handle.

I will send a separate PR to either use `null` for initial `bestBid` and `bestOffer` values or put to market boundaries depending on which is easier.

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
imykolenko: nice name, thx!

PR #164, opened 2024-05-07
CHAIN-145 place initial bestBid/bestOffer on market boundaries

imykolenko:
allows to avoid check for market order being rejected on initial empty market.


bholzman (approved):
Looks good!

PR #165, opened 2024-05-08
CHAIN-149 update markets and symbols while seeding database

imykolenko:


src/main/kotlin/co/chainring/tasks/SeedDatabase.kt
```diff
@@ -35,46 +32,57 @@ fun seedDatabase(fixtures: Fixtures, symbolContractAddresses: List<SymbolContrac
...
-                println("Created market ${marketEntity.guid.value}")
+            MarketEntity.findById(MarketId(baseSymbol, quoteSymbol))
+                ?.also {
+                    it.baseSymbol = baseSymbol
+                    it.quoteSymbol = quoteSymbol
+                    it.tickSize = tickSize
+                    it.flush()
+                    println("Updated market ${it.guid.value}")
+                }
+                ?: run {
```
antonyshchenko, line 71: Not a fan of this style. I think something like 
```
val market = MarketEntity.findById(...)
if (market == null) {
  MarketEntity.create(...)
} else {
   // update market
}
```

is easier to read

antonyshchenko (approved):
Looks good!

src/main/kotlin/co/chainring/tasks/SeedDatabase.kt
imykolenko: update to use`when`, hope it is more readable now

PR #175, opened 2024-05-13
CHAIN-166: 2 phase settlement / withdrawals

bflood366:
The corresponding contracts PR for this feature is here - https://github.com/Chainring-Inc/chainring-contracts/pull/10

- The blockchain transaction handler now manages the 2 phase commit protocol
- it will perform the `prepareBatch` and if all exchanges txs (trade settlements or withdrawals) are good it will invoke the `submitBatch` contract method to submit the batch.
- There is only ever 1 batch is progress at a time (per chain) and currently the next batch is not started till the submit transaction of the current batch has the required number of confirmations.
- If there are any errors in the request to send a prepare or submitBatch transaction to the blockchain (rpc node down or our submitter does not have enough funds for gas), we don't change the batch status but will continue to retry sending the tx to the RPC node.
- IF we successfully send the prepare or submitBatch tx onchain, but it is subsequently reverted, then we mark the batch failed but do not roll anything back or continue to the next batch. We should never revert unless there is a bug, so in this case we stop processing and would need to manually investigate. We would then eventually need to restart the batch manually.
- If individual exchange transactions cannot be applied during the `prepareBatch` phase, events are emitted and the sequencer is invoked to rollback applied changes (only balances are changed). The batch is then re-prepared with the failed txs removed, `prepareBatch` invoked again etc.


sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
```diff
@@ -166,6 +166,37 @@ class SequencerApp(
                     }
                 }
 
+                balanceBatch.failedWithdrawalsList.forEach { failedWithdrawal ->
+                    val wallet = failedWithdrawal.wallet.toWalletAddress()
+                    val asset = failedWithdrawal.asset.toAsset()
+                    val amount = failedWithdrawal.amount.toBigInteger()
+                    state.balances.getOrPut(wallet) { mutableMapOf() }.merge(asset, amount, ::sumBigIntegers)
+                    balancesChanged.merge(Pair(wallet, asset), amount, ::sumBigIntegers)
```
bholzman, line 174: If we reduce a balance, we need to check if there are any orders which need to be auto-reduced

bholzman (approved):
Looks great, just need to add order auto-reduce functionality after a failed withdrawal or trade

backend/src/main/kotlin/co/chainring/core/model/db/ExchangeTransactionBatch.kt
```diff
@@ -0,0 +1,125 @@
...
+    ).index()
+    val error = varchar("error", 10485760).nullable()
+    val sequenceId = integer("sequence_id").autoIncrement()
+    val chainId = reference("chain_id", ChainTable)
+    val prepareBlockchainTransactionGuid = reference(
+        "prepare_blockchain_transaction_guid",
+        BlockchainTransactionTable,
+    )
+    val submitBlockchainTransactionGuid = reference(
+        "submit_blockchain_transaction_guid",
```
antonyshchenko, line 47: Nit: column names could be shorter, e.g `prepare_tx_guid`, `submit_tx_guid`.

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -76,30 +83,108 @@ class BlockchainTransactionHandler(
...
+                                    )?.let {
+                                        Exchange.getPrepareTransactionFailedEventFromLog(eventLog)
+                                    }
+                                }
+                                if (failedTxs.isEmpty()) {
+                                    currentBatch.markAsPrepared()
+                                } else {
+                                    failedTxs.forEach { failedTx ->
+                                        val errorMsg = errorCodeToString(failedTx.errorCode)
+                                        logger.warn { "Received failed event for sequence ${failedTx.sequence}, errorMsg=$errorMsg errorCode=${failedTx.errorCode}" }
```
antonyshchenko, line 118: Should this be an `error` log so that we are notified? Might be helpful at least in the beginning.

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -76,30 +83,108 @@ class BlockchainTransactionHandler(
...
+                }
+            }
+            ExchangeTransactionBatchStatus.Prepared -> {
+                if (currentBatch.submitBlockchainTransaction == null) {
+                    // create blockchain entity and try to submit
+                    createSubmitTransaction(currentBatch)?.let {
+                        submitBatch(currentBatch, it)
+                    }
+                } else {
+                    // this is a retry case - we created the entity but was not able to submit it
```
antonyshchenko, line 149: Would be nice to get an alert after some number of retries

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -292,47 +432,61 @@ class BlockchainTransactionHandler(
...
+
                         if (error != null) {
-                            BlockchainClient.logger.error { "settlement failed for ${tx.tradeId} - error is <$error>" }
-                            tradeEntity.failSettlement()
+                            logger.error { "settlement failed for ${tx.tradeId} - error is <$error>" }
+                            tradeEntity.failSettlement(error)
+                            val buyOrder = executions.first { it.order.side == OrderSide.Buy }.order
+                            val sellOrder = executions.first { it.order.side == OrderSide.Sell }.order
+                            runBlocking {
+                                sequencerClient.failSettlement(
```
antonyshchenko, line 443: What would happen if Sequencer returns a failure here? I would argue that we should check the `success` field in `SequencerResponse` and at least log an error. Same applies for `failWithdraw` case.

sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
bflood366: auto reduce is already happening farther down as part of the BalanceBatch response (it's autoreducing using all the balance change keys - I will add a test for it though

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -76,30 +83,108 @@ class BlockchainTransactionHandler(
...
+                                if (failedTxs.isEmpty()) {
+                                    currentBatch.markAsPrepared()
+                                } else {
+                                    failedTxs.forEach { failedTx ->
+                                        val errorMsg = errorCodeToString(failedTx.errorCode)
+                                        logger.warn { "Received failed event for sequence ${failedTx.sequence}, errorMsg=$errorMsg errorCode=${failedTx.errorCode}" }
+                                        ExchangeTransactionEntity.findForBatchAndSequence(
+                                            currentBatch,
+                                            failedTx.sequence.toInt(),
+                                        )?.let { exchangeTx ->
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
bflood366: changed

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
bflood366: I now log an error if the sequencer returns an error here of for the failed withdrawal

backend/src/main/kotlin/co/chainring/core/model/db/ExchangeTransactionBatch.kt
bflood366: changed

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -76,30 +83,108 @@ class BlockchainTransactionHandler(
...
-        if (uncompletedTransactions.isNotEmpty()) {
-            val currentBlock = blockchainClient.getBlockNumber()
+            else -> {}
+        }
 
-            // Refresh submitted - if we hit required confirmations, invoke callbacks
-            uncompletedTransactions
-                .filter { it.status == BlockchainTransactionStatus.Submitted }
-                .forEach { tx -> refreshSubmittedTransaction(tx, currentBlock) }
+        return true
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -76,30 +83,108 @@ class BlockchainTransactionHandler(
...
+                        submitBatch(currentBatch, it)
+                    }
+                } else {
+                    // this is a retry case - we created the entity but was not able to submit it
+                    if (currentBatch.submitBlockchainTransaction?.status == BlockchainTransactionStatus.Pending) {
+                        submitBatch(currentBatch, currentBatch.submitBlockchainTransaction!!)
+                    }
+                }
+                if (currentBatch.prepareBlockchainTransaction.status == BlockchainTransactionStatus.Pending) {
+                    return false
```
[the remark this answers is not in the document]
bflood366: I changed it but my initial intent here is that this case happens if we are not able to submit the tx onchain, either rpc node is down or the submitter does not have enough funds to pay the gas, so no need to retry so soon.

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
bflood366: the submit logs an error if this happens so we will get an alert I believe

PR #176, opened 2024-05-14
Implement fees

antonyshchenko:
- Fee rates are set in basic points (1 BP = 0.01%)
- Fee rates for maker and taker can differ
- Fee rates are set in Sequencer via RPC
- Sequencer acts as a source of truth for fee rates, backend DB has a copy stored in `key-value` table for UI
- Change of fee rates does not affect the orders that are already in the book
- If existing order is changed, the latest fee rate is applied
- Fees are taken into account when calculating limits

### DIFF
--- a/backend/src/main/kotlin/co/chainring/apps/api/model/Configuration.kt
+++ b/backend/src/main/kotlin/co/chainring/apps/api/model/Configuration.kt
@@ -3,19 +3,14 @@ package co.chainring.apps.api.model
 import co.chainring.core.model.Address
 import co.chainring.core.model.Symbol
 import co.chainring.core.model.db.ChainId
+import co.chainring.core.model.db.FeeRates
 import kotlinx.serialization.Serializable
 
 @Serializable
 data class ConfigurationApiResponse(
     val chains: List<Chain>,
     val markets: List<Market>,
-    val feeRatesInBps: FeeRatesInBps,
-)
-
-@Serializable
-data class FeeRatesInBps(
-    val maker: Int,
-    val taker: Int,
+    val feeRates: FeeRates,
 )
 
 @Serializable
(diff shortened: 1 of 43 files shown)


web-ui/src/components/Screens/HomeScreen/TradeWidget.tsx
```diff
@@ -86,13 +88,24 @@ export default function TradeWidget({
...
+        notional,
+        fee: (notional * feeRatesInBps.taker) / 10000n
+      }
     } else {
-      return (price * amount) / BigInt(Math.pow(10, baseSymbol.decimals))
+      const notional =
+        (price * amount) / BigInt(Math.pow(10, baseSymbol.decimals))
+      return {
+        notional,
+        fee: (notional * feeRatesInBps.maker) / 10000n
```
antonyshchenko, line 105: Realised that this won't be accurate in case order crosses the market. Should I update the calculation here to be more accurate?

bholzman: it's a good point, although any calculation here is purely indicative since the actual price can be different by the time the order is actually placed, so maybe not worth overthinking it. maybe just do it in a separate PR? note, btw, that this widget was renamed to `OrderTicketWidget`

bholzman (approved):
Looks great, nice work!

backend/src/main/kotlin/co/chainring/apps/api/model/Configuration.kt
```diff
@@ -9,6 +9,13 @@ import kotlinx.serialization.Serializable
 data class ConfigurationApiResponse(
     val chains: List<Chain>,
     val markets: List<Market>,
+    val feeRatesInBps: FeeRatesInBps,
+)
+
+@Serializable
+data class FeeRatesInBps(
```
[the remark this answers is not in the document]
antonyshchenko: yup, done

web-ui/src/components/Screens/HomeScreen/TradeWidget.tsx
antonyshchenko: For now I've left it practically as is, with a slight change so that fee is always shown with `~` prefix no matter order type.

PR #184, opened 2024-05-16
Chain 161/fine tune pan/zoom behavior

imykolenko:
- panning
    - no future times on the graph
    - panning via click+horizontal drag
    - panned to mostright point (half of the last ohlc is visible) - graph follows new ohlcs
    - panned left - graph is static

 - zoom
    - zooming via click+vertical drag
    - mostright point is fixed
    - auto switch ohlc duration (less than 20 candles, more that 200 candles), otherwise limit max/min zoom

 - limit number of initial ohlcs provided by BE

### DIFF
--- a/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
@@ -1,6 +1,12 @@
 import { Widget } from 'components/common/Widget'
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { OHLC, OHLCDuration, pricesTopic, Publishable } from 'websocketMessages'
+import {
+  OHLC,
+  OHLCDuration,
+  OHLCDurationSchema,
+  pricesTopic,
+  Publishable
+} from 'websocketMessages'
 import { mergeOHLC, ohlcDurationsMs } from 'utils/pricesUtils'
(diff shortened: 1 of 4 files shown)


web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
```diff
@@ -164,17 +218,67 @@ function OHLCChart({
...
+
+      const newDomain: [Date, Date] = [
+        new Date(currentBeginOfTheWorldTime + autoPanTime),
+        new Date(currentEndOfTheWorldTime + autoPanTime)
+      ]
+
+      domainXRef.current = newDomain
+      drawChart(xScale.domain(newDomain))
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bholzman, line 266: is this safe? I get that domainXRef is dependent on ohlc changing, but seems like autoPanRight could change but this effect would not be updated

bholzman (approved):
Looks great!

web-ui/src/components/Screens/HomeScreen/PricesWidget.tsx
imykolenko: you are right, excluding dependencies is potentially unsafe. Listed all deps, as well changed member definition order and introduced some additional memos and callbacks.

PR #186, opened 2024-05-16
Use shared sign-auth-token promise

imykolenko:
On auth token expiry multiple requests fail with 401 and cause signing of new auth token. Use shared signing promise to deduplicate.


bholzman (approved):
Looks good!

PR #187, opened 2024-05-16
Sequencer test improvements

antonyshchenko:
- made numbers easier to read
- reduced number conversion noise by using BigDecimals in tests and doing BigDecimal -> BigInteger conversion in the test client
- added some helper functions for response verification

### DIFF
--- a/sequencer/src/test/kotlin/co/chainring/testutils/SequencerClient.kt
+++ b/sequencer/src/test/kotlin/co/chainring/testutils/SequencerClient.kt
@@ -12,6 +12,8 @@ import co.chainring.sequencer.core.toDecimalValue
 import co.chainring.sequencer.core.toIntegerValue
 import co.chainring.sequencer.core.toMarketId
 import co.chainring.sequencer.proto.Order
+import co.chainring.sequencer.proto.OrderChanged
+import co.chainring.sequencer.proto.OrderDisposition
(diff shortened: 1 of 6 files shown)


sequencer/src/test/kotlin/co/chainring/testutils/SequencerClient.kt
```diff
@@ -43,51 +66,62 @@ class SequencerClient {
...
+    fun addOrderAndVerifyAccepted(
+        market: Market,
+        amount: BigDecimal,
+        price: BigDecimal?,
+        wallet: WalletAddress,
+        orderType: Order.Type,
+    ): OrderChanged =
+        addOrder(market, amount, price, wallet, orderType).also {
+            assertEquals(OrderDisposition.Accepted, it.ordersChangedList.first().disposition)
+        }.ordersChangedList.first()
```
bholzman, line 92: nit: could move `ordersChangedList.first()` to before the `.also`

bholzman (approved):
Looks great, much cleaner

sequencer/src/test/kotlin/co/chainring/testutils/SequencerClient.kt
antonyshchenko: yup, thx

PR #188, opened 2024-05-17
CHAIN-161 pinch to zoom/pan on mobile devices

imykolenko:
Used `@use-gesture` hook (https://use-gesture.netlify.app/) to handle drag and pinch events on different platforms.


bholzman (approved):
Nice!

PR #192, opened 2024-05-17
CHAIN-189: add support for a 2nd chain

bflood366:
- contains backend / infra changes plus minimal UI changes (to show symbols on all chains plus change to send signing chain for orders). 
- The chain switching for deposits / withdrawals will be done in separate ticket


integrationtests/src/main/kotlin/co/chainring/integrationtests/utils/Wallet.kt
```diff
@@ -31,16 +33,19 @@ class Wallet(
     val walletKeypair: ECKeyPair,
     val contracts: List<DeployedContract>,
     val symbols: List<SymbolInfo>,
+    val chainId: ChainId,
```
bholzman, line 36: I wonder if this is the most natural model for the test wallet, since generally the same wallet would be used on multiple chains. So seems like maybe it would be better to either pass in the chain id on all the individual operations, or perhaps more ergonomically, allow the chain to be changed.

web-ui/.env
```diff
@@ -6,3 +6,10 @@ ENV_NATIVE_SYMBOL_NAME=Bitcoin
 ENV_CHAIN_JSON_RPC_URL=http://localhost:8545
 ENV_CHAIN_BLOCK_EXPLORER_NAME=Otterscan
 ENV_CHAIN_BLOCK_EXPLORER_URL=http://localhost:5100/
+ENV_CHAIN_ID_2=1338
+ENV_CHAIN_NAME_2=Localhost2
+ENV_NATIVE_SYMBOL_2=BTC2
+ENV_NATIVE_SYMBOL_NAME_2=Bitcoin
+ENV_CHAIN_JSON_RPC_URL_2=http://localhost:8546
+ENV_CHAIN_BLOCK_EXPLORER_NAME_2=Otterscan2
+ENV_CHAIN_BLOCK_EXPLORER_URL_2=http://localhost:5101/
```
bholzman, line 15: I wonder if we need to supply any of this via env variables, can we provide it via the config API instead?

scripts/envs/test.yml
```diff
@@ -1,7 +1,9 @@
 environment:
   api:
     DB_HOST: test-db-cluster.cluster-cpwwaa4iqa1s.us-east-2.rds.amazonaws.com
-    EVM_NETWORK_URL: http://test-anvil.chainring.co:8545
+    EVM_CHAINS: anvil,anvil2
+    EVM_NETWORK_URL_anvil: http://test-anvil.chainring.co:8545
+    EVM_NETWORK_URL_anvil2: http://test-anvil2.chainring.co:8545
```
bholzman, line 6: I wonder if it would make more sense to have this config in the DB instead of in environment variables?

bholzman (approved):
Looks good!

README.md
```diff
@@ -69,11 +69,17 @@ make run_backend
    - Put `http://localhost:8545` as RPC URL
    - Put `0x539` as Chain ID
    - Put `ETH` as Currency symbol
-3. Add a local account to Metamask:
+3. Add 2nd local network (served by Anvil2) to Metamask:
```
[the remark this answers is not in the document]
bflood366: yeah good point - I can remove this

integrationtests/src/main/kotlin/co/chainring/integrationtests/utils/Wallet.kt
bflood366: I'll change it to allow the chain to be switched for the wallet

web-ui/.env
bflood366: I'll adda ticket for this

scripts/envs/test.yml
bflood366: and for this

integrationtests/src/main/kotlin/co/chainring/integrationtests/utils/Wallet.kt
bflood366: changed so same wallet can be used across chains - wallet contains a `switchChain` to switch between chains

web-ui/.env
bflood366: Added https://chainringinc.atlassian.net/browse/CHAIN-207 for this and issue below

PR #211, opened 2024-05-24
Added websocket reconnect logic to mock maker and mock taker

antonyshchenko:
- had to switch to non-blocking websockets since blocking ones can't detect broken connection and just block forever
- switched to websocket client from OkHTTP because non-blocking client from http4k does not support reconnecting
- did some refactoring and cleanup


mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
```diff
@@ -23,182 +23,142 @@ import co.chainring.core.model.db.OrderSide
...
+                            logger.info { "$id: order ${order.id} fully filled ${order.amount}" }
+                            currentOrders[order.marketId]!!.removeIf { it.id == order.id }
+                        }
 
-                                    else -> {}
-                                }
-                            }
+                        OrderStatus.Cancelled -> {
+                            logger.info { "$id: order ${order.id} canceled" }
+                            currentOrders[order.marketId]!!.removeIf { it.id == order.id }
```
imykolenko, line 131: orders are overwritten from api responce every time after submitting batch, this can be skipped

mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
```diff
@@ -311,49 +269,52 @@ class Maker(
...
+
+            val (offerPrices, bidPrices) = offerAndBidPrices(market, levels, curPrice)
+            val (offerAmounts, bidAmounts) = offerAndBidAmounts(market, offerPrices, bidPrices)
+
             offerPrices.forEachIndexed { ix, price ->
                 val amount = offerAmounts[ix]
-                apiClient.tryCreateOrder(
-                    CreateOrderApiRequest.Limit(
+                apiClient.createOrder(
+                    wallet.signOrder(CreateOrderApiRequest.Limit(
```
imykolenko: create order throws exception on any unsuccessful api call, it will skip other orders from being created. Can exception affect websocket connectivity?

imykolenko (approved):
Nice cleanup 👍

mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
antonyshchenko: Changed it to `tryCreateOrder`

mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
antonyshchenko: Leaving it as is just in case

PR #214, opened 2024-05-24
Limit the size of orders list sent via WS to prevent OOM errors in API

antonyshchenko:


backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
```diff
@@ -279,17 +280,21 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
...
-            if (statuses.isNotEmpty()) {
-                query = query.and(OrderTable.status.inList(statuses))
-            }
+        fun listForWallet(wallet: WalletEntity, statuses: List<OrderStatus> = emptyList(), marketId: MarketId? = null, limit: Int? = null): List<OrderEntity> {
             return OrderEntity
-                .find(query)
+                .find(
+                    OrderTable.walletGuid.eq(wallet.guid)
+                        .andIfNotNull(marketId?.let { OrderTable.marketGuid.eq(it) })
+                        .andIfNotNull(statuses.ifEmpty { null }?.let { OrderTable.status.inList(statuses) }),
```
imykolenko, line 288: for MMs there are tens of thousands of orders. To get this filter executed efficiently it looks like few more multicolumn indexes need to be added (`wallet_guid + status` and `wallet_guid + market_guid + status`

btw, nice use of  `andIfNotNull`

imykolenko (approved):
looks great

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
antonyshchenko: Added a [ticket](https://chainringinc.atlassian.net/browse/CHAIN-242) to address this and the N+1 query problem

PR #217, opened 2024-05-27
demo env faucet

imykolenko:
- only native tokens can be selected on all available chains
 - amount is fixed to 0.1
 - connected wallet's address is used

======
Screenshot 2024-05-27 at 11 57 11 AM
Screenshot 2024-05-27 at 11 57 23 AM
Screenshot 2024-05-27 at 11 57 31 AM


docker-compose.yaml
```diff
@@ -38,6 +36,11 @@ services:
...
     image: postgres:15
     restart: always
+    volumes:
+      - type: tmpfs
+        target: /var/lib/postgresql/data:rw
+        tmpfs:
+          size: 4294967296
     environment:
       POSTGRES_USER: chainring
       POSTGRES_PASSWORD: chainring
```
imykolenko, line 46: this seems to be working to overcome my local issue when postgres image restarts claiming disk is full

antonyshchenko: Nice! 👍

web-ui/src/components/Screens/Header.tsx
```diff
@@ -81,6 +85,28 @@ export function Header({
     )
   }
 
+  function faucet(onOpen: () => void) {
```
antonyshchenko, line 88: Nit: rename to `faucetButton` and make it take object as an argument so that we could write something like this:

```
faucetButton({onOpen: () => { setShowFaucetModal(true) })
```

web-ui/src/components/Screens/HomeScreen/faucet/FaucetModal.tsx
```diff
@@ -0,0 +1,176 @@
...
+          </div>
+          <div className="mt-2">
+            <div className="mr-2 text-darkBluishGray1">Amount:</div>
+            <div>0.1 {selectedSymbol?.name}</div>
+          </div>
+
+          {phase == 'sent' ? (
+            <div className="mt-2 flex text-center text-statusGreen">
+              Your request has been processed. The tokens may arrive with a
+              slight delay.
```
antonyshchenko, line 128: Not sure about `slight delay` part. How about something like this:

```
Success. Your wallet balance will be updated shortly.
```

antonyshchenko (approved):
Looks great! How about adding an integration test for the faucet API endpoint?

web-ui/src/components/Screens/Header.tsx
imykolenko: renamed

web-ui/src/components/Screens/HomeScreen/faucet/FaucetModal.tsx
imykolenko: nice, thx!

imykolenko:
Also added integration test for faucet endpoint.

PR #222, opened 2024-05-28
notify only about affected orders via ws

imykolenko:


backend/src/main/kotlin/co/chainring/apps/api/services/ExchangeApiService.kt
```diff
@@ -328,21 +325,6 @@ class ExchangeApiService(
...
-                            Orders(
-                                OrderEntity
-                                    .listForWallet(walletEntity)
-                                    .map(OrderEntity::toOrderResponse),
-                            ),
-                            walletEntity.address,
-                        ),
-                    ),
-                )
-            }
```
imykolenko, line 345: I also remove pushing all orders to websocket in case when no orders were found on cancel all.

sequencer/src/main/kotlin/co/chainring/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -234,30 +231,16 @@ object SequencerResponseProcessorService {
...
-                BroadcasterNotification(
-                    Orders(
-                        OrderEntity
-                            .listForWallet(walletEntity)
-                            .map(OrderEntity::toOrderResponse),
-                    ),
-                    recipient = walletEntity.address,
-                ),
-            )
-        }
```
imykolenko, line 260: MMs have 650k cancelled orders each and some k filled. When calling cancel-all api endpoint we reached a limit of how big jsonb field can be. 

I replaced `orders` notification to one-by-one `orderUpdated` messages for affected orders.

bflood366 (approved):
looks great

PR #225, opened 2024-05-28
Make Sequencer response processor tolerate database downtime for up to 10 minutes

antonyshchenko:
Retry response processing for up to 10 minutes logging warning on each retry. If error persists past that time log an error and shutdown the entire process gracefully.


sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerResponseProcessorApp.kt
```diff
@@ -111,13 +85,40 @@ class SequencerResponseProcessorApp(
         logger.info { "Stopped" }
     }
 
+    private fun processResponseWithRetries(requestBytes: ByteArray, response: SequencerResponse, lastReadIndex: Long) {
+        val startedAt = Clock.System.now()
+        val maxDuration = 10.minutes
```
bholzman, line 90: Is the idea of the max duration that we would get an alert if that much time passed and it hadn't reconnected? Otherwise what's the harm in just retrying forever? We can obviously stop and restart manually if need be. We could just have the logic here drop a (single) error log if it has been retrying for more than a trivial amount of time (like say 5 seconds)

bholzman (approved):

PR #230, opened 2024-05-29
Reuse balances on maker restart

imykolenko:
- reuse available maker's balance on restart
 - track order state using websocket messages

### DIFF
--- a/mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
+++ b/mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
@@ -45,7 +45,7 @@ class Maker(
 ) : Actor(marketIds, nativeAssets, assets, keyPair) {
     override val id: String = "mm_${Address(Keys.toChecksumAddress("0x" + Keys.getAddress(keyPair))).value}"
     override val logger: KLogger = KotlinLogging.logger {}
-    private var currentOrders = mutableMapOf<MarketId,MutableList<Order.Limit>>()
+    private var currentOrders = mutableMapOf<MarketId,MutableSet<Order.Limit>>()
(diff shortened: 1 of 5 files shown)


mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
```diff
@@ -203,56 +209,58 @@ class Maker(
...
         }.onRight {
-            logger.info { "$id: applied update batch" }
+            logger.info { "$id: applied update batch: created ${createOrders.size}, cancelled ${cancelOrders.size}" }
         }
-
-        currentOrders[marketId] = apiClient
-            .listOrders(listOf(OrderStatus.Open, OrderStatus.Partial), marketId)
-            .orders
-            .map { it as Order.Limit }
-            .toMutableList()
```
imykolenko, line 255: Orders stored to DB async from sequencer response. List orders executed immediately after posting batch returns outdated information, therefore switching to ws updates

mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
```diff
@@ -297,16 +305,8 @@ class Maker(
                 }
             }
 
-            currentOrders[marketId] =
-                apiClient
-                    .listOrders(listOf(OrderStatus.Open, OrderStatus.Partial), marketId)
-                    .orders
-                    .map { it as Order.Limit }
-                    .toMutableList()
```
imykolenko, line 305: same here, after sending 20 createOrder api requests listOrders returns only 8-9 back, rest is still in processing

bflood366 (approved):
looks great 👍

PR #231, opened 2024-05-29
CHAIN-201 fill missing ohlc intervals until current time

imykolenko:
Schedule creation of empty ohlc record (using last close price) exactly on duration change. Data provided by BE replaces any existing data on client side.


bholzman (approved):
Looks good!

PR #234, opened 2024-05-29
Durable Telegram bot sessions

antonyshchenko:
- Session state is stored in `telegram_bot_user` table
- Bot app restores sessions from DB on start
- Callbacks are now handled explicitly (`onCallbackQuery` handler in `Bot`) because `InlineKeyboardItem.onClick` used in prev implementation is storing handlers in memory and they are lost after restart
- Most of message handling and message sending logic is now localized in `BotSession` class


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V34_AddSessionStateToTelegramBotUser.kt
```diff
@@ -0,0 +1,48 @@
...
+        DepositQuoteAmount,
+        WithdrawBaseAmount,
+        WithdrawQuoteAmount,
+    }
+
+    object V34_TelegramBotUserTable : GUIDTable<TelegramBotUserId>("telegram_bot_user", ::TelegramBotUserId) {
+        val createdAt = timestamp("created_at")
+        val createdBy = varchar("created_by", 10485760)
+        val telegramUserId = long("telegram_user_id").uniqueIndex()
+        val currentMarketGuid = reference("current_market_guid", MarketTable).nullable()
```
bholzman, line 31: add a `V34_MarketTable` stub (and use `V34_TelegramUserReplyType` for `expectedReplyType`)

bholzman (approved):
Looks great!

bflood366 (approved):
looks great 👍

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V34_AddSessionStateToTelegramBotUser.kt
antonyshchenko: Right, thanks

PR #235, opened 2024-05-29
CHAIN-245: swap tab

bholzman:
Screenshot 2024-05-29 at 1 43 48 PM
Screenshot 2024-05-29 at 1 43 56 PM
Screenshot 2024-05-29 at 1 44 19 PM
Screenshot 2024-05-29 at 1 44 51 PM
Screenshot 2024-05-29 at 1 44 56 PM
Screenshot 2024-05-29 at 1 44 59 PM


bflood366:
a few questions on the screen shots:
- in the first screen you are selling 1.5, so I would expect knowing the fee would be 0.03 (assuming the rate is 200bips) but it looks like .02999. Similar for the screen where you are buying 1.5 ETH its 0.02986..
- the last screens where you don't have a sufficient BTC balance - it looks like you need 108 and you have 107.17. on deposit  I guess I would have expected 0.83 would be auto-populated as the amount needed since this is base symbol. Was there another order to sell 1BTC so the limit was really 106.17. if so it seems a little confusing to show the amount on deposit right there since mentally that is what I think I have available for the order, but I really only have what my limit is.

bflood366 (approved):
looks great to me. nice job 👍

PR #236, opened 2024-05-29
CHAIN-177: add otterscan customization

bholzman:


scripts/envs/demo.yml
```diff
@@ -57,7 +61,7 @@ services:
     count: 1
     is_essential: "true"
   otterscan:
-    image: otterscan/otterscan
+    image: 851725450525.dkr.ecr.us-east-2.amazonaws.com/otterscan
```
bflood366, line 64: are there terraform changes needed for this ecr registry.

bflood366 (approved):
looks great

imykolenko (approved):
nice, customized block explorer!
otterscan's docker registry was also very slow.

scripts/envs/demo.yml
bholzman: Indeed there are. I made them, and also refactored the code there a bit to make an `ecr_repo` module.

PR #239, opened 2024-05-30
CHAIN-250: update icons to include chain name

bholzman:
Also added the chain name in text under the symbol name in the symbol selector on the swap screen (but not yet on the Dashboard)

Screenshot 2024-05-29 at 11 59 54 PM
Screenshot 2024-05-29 at 11 59 58 PM
Screenshot 2024-05-30 at 12 00 08 AM


scripts/envs/demo.yml
```diff
@@ -28,7 +28,7 @@ environment:
     NATIVE_CURRENCY_NAME: "Demo Botanix BTC"
   mocker:
     API_URL: http://demo-api.chainring.co:9000
-    EVM_CHAINS: anvil,anvil2
+    EVM_CHAINS: Bitlayer,Botanix
     EVM_NETWORK_URL_anvil: http://demo-anvil.chainring.co:8545
     EVM_NETWORK_URL_anvil2: http://demo-anvil2.chainring.co:8545
```
imykolenko, line 33: anvil/anvil2 also needs to be replaced

web-ui/src/components/common/SubmitButton.tsx
```diff
@@ -4,12 +4,14 @@ export default function SubmitButton({
...
   caption,
-  error
+  error,
+  success
 }: {
   disabled: boolean
   onClick: () => void
   caption: () => string
   error: string | null | undefined
+  success: boolean
```
imykolenko, line 14: It is not immediately clear what `success={false}` passed to a button component means. What if it is renamed to appearance with values `'green' | 'blue'`?

imykolenko (approved):

web-ui/src/components/common/SubmitButton.tsx
antonyshchenko: Or perhaps we could use the presence or absence of `error` to defer the `success` state.

bholzman: Nice, I like that, will do!

scripts/envs/demo.yml
bholzman: Oops, thanks!

PR #255, opened 2024-05-31
fix issue with settlement rollback when some trades succeed and need …

bflood366:
…to be resubmitted


antonyshchenko (approved):
Looks great!

PR #256, opened 2024-06-03
CHAIN-267: fix sequencer crash due to off-by-one error in findNextBes…

bholzman:
…t{Bid,Offer}

Still needs a test.

### DIFF
--- a/sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
+++ b/sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
@@ -506,10 +506,11 @@ data class Market(
 
     private fun findNextBestOffer(level: OrderBookLevel): BigDecimal {
         var index = level.levelIx
-        while (index++ <= maxOfferIx) {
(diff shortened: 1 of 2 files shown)


sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
```diff
@@ -506,21 +506,23 @@ data class Market(
 
     private fun findNextBestOffer(level: OrderBookLevel): BigDecimal {
         var index = level.levelIx
-        while (index++ <= maxOfferIx) {
+        while (index <= maxOfferIx) {
```
bflood366, line 509: I think the `findMinBidIx` and `findMaxOfferIx` should also be adjusted since they also do a `index++` in the while like this.

bholzman: Thanks, adjusted

bflood366 (approved):
looks great

PR #264, opened 2024-06-04
CHAIN-221 fix holding page CORS error

imykolenko:
AWS ALB `FixedResponseConfig` does not support setting custom headers. Plus for CORS preflight request to work server should reply with 200. 

Added simple lambda function using TF and linkEd it to a target group.

=======
Screenshot 2024-06-04 at 3 28 04 PM


terraform/test/main.tf
```diff
@@ -218,10 +218,16 @@ module "sequencer" {
...
-  bastion_ip  = module.bastion.private_ip
 }
+
+#module "loadtest" {
+#  source      = "../modules/loadtest"
+#  name_prefix = local.name_prefix
+#  subnet_id   = module.vpc.private_subnet_id_2
+#  vpc         = module.vpc.vpc
+#  bastion_ip  = module.bastion.private_ip
+#}
```
imykolenko, line 233: removing loadtest ec2 instance for now, it consumes resources.

bholzman (approved):
Nice, good solution for limitations of LB static response

PR #265, opened 2024-06-04
fix flaky tests

bflood366:


integrationtests/src/test/kotlin/co/chainring/integrationtests/testutils/Waiter.kt
```diff
@@ -0,0 +1,68 @@
...
+        .pollDelay(Duration.ofMillis(100))
+        .pollInterval(Duration.ofMillis(100))
+        .atMost(Duration.ofMillis(atMost ?: 30000L))
+        .until {
+            condition()
+        }
+}
+
+private val logger = KotlinLogging.logger { }
+fun waitForActivityToComplete() {
```
bflood366, line 33: this waits now before a test for a finite time for any in progress settlements, deposits or withdrawals to finish

integrationtests/src/test/kotlin/co/chainring/integrationtests/testutils/BalanceHelper.kt
```diff
@@ -62,3 +56,15 @@ fun waitForFinalizedWithdrawal(id: WithdrawalId) {
         }
     }
 }
+
+fun deposit(wallet: Wallet, apiClient: TestApiClient, assetAmount: AssetAmount): TransactionReceipt {
+    return wallet.deposit(assetAmount).also {
```
bholzman, line 61: are there places which call `wallet.deposit()` where we wouldn't want to call `apiClient.createDeposit`? might make more sense to simply have `wallet.deposit()` do this instead of a static helper method?

bholzman (approved):
Looks good!

PR #266, opened 2024-06-04
Use official client library in Telegram bot

antonyshchenko:
I've abstracted the Telegram client into a simple interface so that bot implementation does not depend on a specific library

### DIFF
--- a/telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
+++ b/telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
@@ -2,6 +2,7 @@ package co.chainring.telegrambot.app
 
 import co.chainring.core.model.db.TelegramMessageId
 import co.chainring.core.model.db.TelegramUserId
+import io.github.oshai.kotlinlogging.KotlinLogging
 import org.telegram.telegrambots.client.okhttp.OkHttpTelegramClient
 import org.telegram.telegrambots.longpolling.TelegramBotsLongPollingApplication
 import org.telegram.telegrambots.longpolling.util.LongPollingSingleThreadUpdateConsumer
@@ -16,6 +17,7 @@ import org.telegram.telegrambots.meta.api.objects.replykeyboard.buttons.InlineKe
 class BotTelegramClient(private val token: String) : Bot.TelegramClient {
     private val app = TelegramBotsLongPollingApplication()
     private val telegramClient = OkHttpTelegramClient(token)
+    private val logger = KotlinLogging.logger { }
 
     override fun startPolling(inputHandler: (Bot.Input) -> Unit) {
         app.registerBot(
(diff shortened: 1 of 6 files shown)


telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.telegram.telegrambots.meta.api.methods.send.SendMessage
+import org.telegram.telegrambots.meta.api.methods.updatingmessages.DeleteMessage
+import org.telegram.telegrambots.meta.api.objects.Update
+import org.telegram.telegrambots.meta.api.objects.replykeyboard.ForceReplyKeyboard
+import org.telegram.telegrambots.meta.api.objects.replykeyboard.InlineKeyboardMarkup
+import org.telegram.telegrambots.meta.api.objects.replykeyboard.buttons.InlineKeyboardButton
+import org.telegram.telegrambots.meta.api.objects.replykeyboard.buttons.InlineKeyboardRow
+
+class BotTelegramClient(private val token: String) : Bot.TelegramClient {
+    private val app = TelegramBotsLongPollingApplication()
```
bflood366, line 17: not sure if its needed but is there any wire level logging to aid in debugging or is there anything we would want to log

telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
```diff
@@ -0,0 +1,113 @@
...
+        app.registerBot(
+            token,
+            object : LongPollingSingleThreadUpdateConsumer {
+                override fun consume(update: Update) {
+                    val message = update.message
+                    val callbackQuery = update.callbackQuery
+
+                    if (message != null && message.hasText()) {
+                        if (message.text == "/start") {
+                            inputHandler(
```
bflood366, line 30: do we know what would happen if the inputHandler threw an unexpected exception. Would it cause this consume thread to exist or does `LongPollingSingleThreadUpdateConsumer` catch it and keep going and if that did happen would we see anything logged anywhere

bflood366 (approved):
looks great 👍

imykolenko (approved):
nice!

telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
antonyshchenko: Added debug logging for input and output messages, although we should turn it off on prod in order not to leak private keys

telegrambot/src/main/kotlin/co/chainring/telegrambot/app/BotTelegramClient.kt
antonyshchenko: I've checked and it does not exit or crash on unhandled exceptions. However I've wrapped handling in try/catch anyways in order to log the error and send `Something went wrong, please try again later` back to the user.

PR #271, opened 2024-06-05
Swap interface in Telegram bot

antonyshchenko:
- bot is now modelled as a finite state machine
- pending airdrops, deposits, withdrawals and orders are tracked by a separate thread w/o relying on websockets

Created separate tickets for:
- market price should be calculated based on the order book and amount, similar to how it is done in web UI
- should consider limits and fees before allowing to submit an order

https://github.com/Chainring-Inc/chainring-monorepo/assets/443906/c76b0742-01d6-4aff-87be-6dd5c650e559

### DIFF
--- a/backend/src/main/kotlin/co/chainring/core/db/Database.kt
+++ b/backend/src/main/kotlin/co/chainring/core/db/Database.kt
@@ -3,10 +3,15 @@ package co.chainring.core.db
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 import org.apache.commons.dbcp2.BasicDataSource
+import org.jetbrains.exposed.dao.Entity
+import org.jetbrains.exposed.dao.EntityClass
 import org.jetbrains.exposed.sql.Database
 import org.jetbrains.exposed.sql.DatabaseConfig
+import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.Transaction
+import org.jetbrains.exposed.sql.transactions.TransactionManager
 import java.lang.System.getenv
+import java.sql.ResultSet
 import java.time.Duration
 
 @Serializable
@@ -73,3 +78,16 @@ fun Transaction.notifyDbListener(
         ).joinToString(", "),
     )
 }
+
+fun <T : Entity<Q>, Q : Comparable<Q>> Transaction.executeRaw(sql: String, entity: EntityClass<Q, T>): List<T> {
+    val result = arrayListOf<T>()
+    TransactionManager.current().exec(sql) { rs ->
+        while (rs.next()) {
+            result += entityFromResultSet(rs, entity)
+        }
+    }
+    return result
+}
+
+fun <T : Entity<Q>, Q : Comparable<Q>> Transaction.entityFromResultSet(rs: ResultSet, entity: EntityClass<Q, T>): T =
+    entity.wrapRow(ResultRow.create(rs, entity.table.columns.mapIndexed { index, field -> field to index }.toMap()))
--- a/backend/src/main/kotlin/co/chainring/core/db/Migrations.kt
+++ b/backend/src/main/kotlin/co/chainring/core/db/Migrations.kt
@@ -36,6 +36,7 @@ import co.chainring.core.model.db.migrations.V3_UpdateDeployedSmartContract
 import co.chainring.core.model.db.migrations.V40_OrderIndexes
 import co.chainring.core.model.db.migrations.V41_AddActualAmountToWithdrawalTable
 import co.chainring.core.model.db.migrations.V42_ChainConfiguration
+import co.chainring.core.model.db.migrations.V43_AddFsmSessionStateToTelegramBotUser
 import co.chainring.core.model.db.migrations.V4_AddDecimalsToERC20Token
 import co.chainring.core.model.db.migrations.V5_ChainTable
 import co.chainring.core.model.db.migrations.V6_MarketTable
@@ -86,4 +87,5 @@ val migrations = listOf(
     V40_OrderIndexes(),
     V41_AddActualAmountToWithdrawalTable(),
     V42_ChainConfiguration(),
+    V43_AddFsmSessionStateToTelegramBotUser(),
 )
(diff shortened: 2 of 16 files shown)


bholzman:
> This is more of a draft, to get some feedback on the interface. There are some pieces missing:
> 
> * market price should be calculated based on the order book and amount, similar to how it is done in web UI
> * should consider limits and fees before allowing to submit an order
> 
>  tg-swap-s.mov

So I think we should completely remove the concept of "Market" from here. What if it was like this -- there's just a button "Swap" which when you click it there's a list of assets to sell; when they click one, there's a list of assets to buy, and then they just have to give the amount to do the swap. Then, we could remember their "sell token" and "buy token" and give them a "Swap more BTC for ETH" (or whatever) button.

antonyshchenko:
I like that @bholzman. I will give it a try tomorrow. Here is an example how I see the flow:

```
1. User enters /start command
2. Bot shows main menu with wallet address, balances and buttons: [Swap] [Deposit] [Withdraw] [Settings]
3. User taps [Swap]
4. Bot: Select token to sell: [BTC] [ETH] ... [USDC] [Cancel]
5. User taps [BTC]
6. Bot: Select token to sell: [ETH] ... [USDC] [Cancel]
7. User taps [ETH] 
8. Bot: You have X BTC. How much are you willing to swap?
9. User enters amount X
10. Bot: Selling X BTC will give you Y ETH at the price of Z ETH with a fee of Q ETH. Proceed? [Confirm] [Cancel] [Change amount]
11. User taps [Confirm]
12. Bot: Order submitted
13. Bot: Order filled
14. Bot: Trade settled
15. Bot: shows main menu with balances and buttons: [Swap more BTC for ETH] [Swap ETH for BTC] [Swap other tokens] [Deposit] [Withdraw] ...
```

I'm not sure how step 8 and 10 should look like when user is willing to sell quote symbol, say ETH in the BTC/ETH market. We have to consider that fee is paid in quote currency, so we can't just say `You have X ETH. How much are you willing to swap?` Or perhaps we can, but should reject the input with an explanation if what they have entered + fee is more than what they have. Any thoughts on this?

bholzman:
> I like that @bholzman. I will give it a try tomorrow. Here is an example how I see the flow:
> 
> ```
> 1. User enters /start command
> 2. Bot shows main menu with wallet address, balances and buttons: [Swap] [Deposit] [Withdraw] [Settings]
> 3. User taps [Swap]
> 4. Bot: Select token to sell: [BTC] [ETH] ... [USDC] [Cancel]
> 5. User taps [BTC]
> 6. Bot: Select token to sell: [ETH] ... [USDC] [Cancel]
> 7. User taps [ETH] 
> 8. Bot: You have X BTC. How much are you willing to swap?
> 9. User enters amount X
> 10. Bot: Selling X BTC will give you Y ETH at the price of Z ETH with a fee of Q ETH. Proceed? [Confirm] [Cancel] [Change amount]
> 11. User taps [Confirm]
> 12. Bot: Order submitted
> 13. Bot: Order filled
> 14. Bot: Trade settled
> 15. Bot: shows main menu with balances and buttons: [Swap more BTC for ETH] [Swap ETH for BTC] [Swap other tokens] [Deposit] [Withdraw] ...
> ```
> 
> I'm not sure how step 8 and 10 should look like when user is willing to sell quote symbol, say ETH in the BTC/ETH market. We have to consider that fee is paid in quote currency, so we can't just say `You have X ETH. How much are you willing to swap?` Or perhaps we can, but should reject the input with an explanation if what they have entered + fee is more than what they have. Any thoughts on this?

I think probably a "Sell Max" button would handle it most elegantly. Obviously wait until we've implemented that in the sequencer to do it here. We could show an error as you suggest, I think that makes sense.

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V43_AddSessionStateToTelegramBotUser.kt
```diff
@@ -0,0 +1,50 @@
...
+        val messageIdsForDeletion = array<Int>("message_ids_for_deletion").default(emptyList())
+        val sessionState = jsonb<V43_BotSessionState>("session_state", KotlinxSerialization.json).default(V43_BotSessionState.Initial)
+    }
+
+    override fun run() {
+        transaction {
+            exec("ALTER TABLE telegram_bot_user ADD COLUMN updated_at timestamp without time zone")
+            exec("UPDATE telegram_bot_user SET updated_at = now()")
+            exec("ALTER TABLE telegram_bot_user ALTER COLUMN updated_at SET NOT NULL")
+            SchemaUtils.createMissingTablesAndColumns(V43_TelegramBotUserTable)
```
bholzman, line 41: fwiw couldn't this be first and then we wouldn't need the "ADD COLUMN updated_at" line?

bholzman (approved):
Looks great! FSM model makes it very easy to follow.

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V43_AddSessionStateToTelegramBotUser.kt
antonyshchenko: The column is non nullable, so it has to be done in 3 steps:
- add a nullable column
- set value for existing records
- change it to non nullable

bholzman: Yes I know, but you defined it as nullable in the table here so you could use schemautils to create it and then the two commands to update and make it not nullable

antonyshchenko: Here is how it is defined in the migration:
```
 val updatedAt = timestamp("updated_at")
```

Defining it as nullable would look like this:
```
 val updatedAt = timestamp("updated_at").nullable()
```

bholzman: True, true. I guess I’ve usually done it that way, but it doesn’t really make a difference. Nevermind.

PR #276, opened 2024-06-06
Withdraw all functionality and fail withdraw if not sufficient balance

bflood366:


backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -370,17 +378,37 @@ class BlockchainTransactionHandler(
...
+    private fun completeWithdrawals(settlingWithdrawals: List<WithdrawalEntity>, withdrawalEvents: Map<Int, Exchange.WithdrawalEventResponse> = emptyMap()) {
         settlingWithdrawals.forEach {
             try {
-                onTxComplete(it.transactionData!!, error)
+                val withdrawAmount = withdrawalEvents[it.sequenceId]?.let { event ->
+                    if (it.amount == BigInteger.ZERO && it.actualAmount != null && event.amount.compareTo(it.actualAmount) != 0) {
+                        logger.warn { "Withdraw all mismatch, sequencer=${it.actualAmount} contract=${event.amount} " }
+                    }
+                    event.amount
+                } ?: it.actualAmount ?: it.amount
```
bholzman, line 396: under what conditions would there not be an event? would it be better to log an error and not complete the withdrawal in that case?

backend/src/main/kotlin/co/chainring/core/db/Migrations.kt
```diff
@@ -76,4 +77,5 @@ val migrations = listOf(
     V35_RemoveExchangeContract(),
     V36_AddPendingDepositsToTelegramBotUser(),
     V37_AddBatchHash(),
+    V38_AddActualAmountToWithdrawalTable(),
```
bholzman, line 80: fyi I just merged a change that added a V38_ migration

bholzman (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
bflood366: yeah there should always be an event. I will adjust this - it was primarily defensive, but better to fail fast if this happens and alert.

backend/src/main/kotlin/co/chainring/core/db/Migrations.kt
bflood366: thanks will adjust - also there is a contract PR with this - so can't merge til that is reviewed - https://github.com/Chainring-Inc/chainring-contracts/pull/15

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
```diff
@@ -69,7 +70,13 @@ sealed class EIP712Transaction {
 
         override fun getTxData(sequence: Long): ByteArray {
             return serializeTx(
-                ExchangeTransactions.WithdrawWithSignature(ExchangeTransactions.Withdraw(sequence, sender.value, token.address.value, amount, nonce.toBigInteger()), signature.toByteArray()),
+                ExchangeTransactions.WithdrawWithSignature(ExchangeTransactions.Withdraw(sequence, sender.value, token.address.value, amount, nonce.toBigInteger()), signature.toByteArray(), BigInteger.ZERO),
+            )
+        }
+
+        fun getTxData(sequence: Long, withdrawAllAmount: BigInteger = BigInteger.ZERO): ByteArray {
```
antonyshchenko, line 77: I don't quite understand why do we need `withdrawAllAmount` to be honest. It is rather confusing tbh to have 2 amount params in the `withdraw` function of the contract. I've left a similar comment in the contract PR.

bflood366: I commented in the contracts PR what the rationale was for this approach

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
bflood366: Actually there is a case where there might not an event albeit rare - if anvil restarts while a withdrawal batch is in progress we do not get a receipt and the code assumes all transactions in the batch were successful - for now I changed to log if we do not get an event, but don't stop or fail anything,

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -226,28 +225,34 @@ class BlockchainTransactionHandler(
...
-                                settlingWithdrawals.firstOrNull { it.sequenceId.toBigInteger() == failedWithdrawal.sequence }?.let {
-                                    completeWithdrawals(listOf(it), errorMsg)
+                            }.toMap()
+                            val completedWithdrawalBySequence = txReceipt.logs.mapNotNull { eventLog ->
+                                Contract.staticExtractEventParameters(
+                                    Exchange.WITHDRAWAL_EVENT,
+                                    eventLog,
+                                )?.let {
+                                    val event = Exchange.getWithdrawalEventFromLog(eventLog)
+                                    Pair(event.sequence.toInt(), event)
```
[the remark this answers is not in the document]
bflood366: changed to a long

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
bflood366: I updated the contracts and adjusted here to not send a separate withdrawAllAmount value.

PR #277, opened 2024-06-06
optimize order listing

bflood366:


backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
```diff
@@ -282,13 +297,25 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
...
+
+        private fun listOrders(orderIds: List<OrderId>): List<OrderEntity> {
             return OrderEntity
                 .find { OrderTable.guid.inList(orderIds) }
                 .orderBy(Pair(OrderTable.createdAt, SortOrder.DESC))
                 .toList()
         }
 
+        fun listOrdersWithExecutions(orderIds: List<OrderId>): List<Pair<OrderEntity, List<OrderExecutionEntity>>> {
+            val orders = listOrders(orderIds)
```
bholzman, line 314: would it be better to use a single query with joins instead?

bflood366: changed to a single query with a left join.

bholzman (approved):
Looks good!

PR #280, opened 2024-06-10
Feature/chain 207 env chain info via conf

imykolenko:
- load chains configuration before rendering the app
 - fixtures include local config, I will populate `rpc_url`s and `block_explorer_url`s for other envs manually on deployment


web-ui/src/index.tsx
```diff
@@ -1,18 +1,53 @@
...
+          </span>
+        </div>
+      )}
+      {wagmiConfig ? (
+        <WagmiProvider config={wagmiConfig}>
+          <QueryClientProvider client={queryClient}>
+            <App />
+          </QueryClientProvider>
+        </WagmiProvider>
+      ) : (
```
imykolenko, line 44: Wagmi provider enforces us to provide config with non-empty chains, therefore loading configuration from BE as a very first step before rendering the app.

gradle.properties
```diff
@@ -1,2 +1,3 @@
 kotlin.code.style=official
-kotlin.daemon.jvmargs=-Xmx2048m
\ No newline at end of file
+kotlin.daemon.jvmargs=-Xmx2048m
+kotlin.compiler.execution.strategy=in-process
```
bholzman, line 3: I guess this helps with the issue of not being able to reuse the kotlin daemon? Wondering if we know the root cause of that?

web-ui/src/wagmiConfig.ts
```diff
@@ -12,76 +13,76 @@ const walletConnectMetadata = {
...
-    default: { http: [import.meta.env.ENV_CHAIN_JSON_RPC_URL] }
-  },
-  blockExplorers: {
-    default: {
-      name: import.meta.env.ENV_CHAIN_BLOCK_EXPLORER_NAME,
-      url: import.meta.env.ENV_CHAIN_BLOCK_EXPLORER_URL
-    }
-  }
-})
+export let allChains: [Chain, ...Chain[]]
```
bholzman, line 16: this is cool

web-ui/src/wagmiConfig.ts
```diff
@@ -12,76 +13,76 @@ const walletConnectMetadata = {
...
-    }),
-    injected({ shimDisconnect: true }),
-    coinbaseWallet({
-      appName: walletConnectMetadata.name,
-      appLogoUrl: walletConnectMetadata.icons[0]
+    return defineChain({
+      id: chain.id,
+      name: chain.name,
+      nativeCurrency: {
+        decimals: 18,
```
bholzman, line 33: probably this is always fine, but couldn't we just read it from the native symbol?

bholzman (approved):
Looks good!

gradle.properties
imykolenko: Correct, this will start compiler in process without trying to connect to daemon. Don't know what was the reason, some incompatibility. On CI instances daemon is never there, might have slight memory effect locally though, I'll drop it.

web-ui/src/wagmiConfig.ts
imykolenko: yep! typed definition of non-empty array

web-ui/src/wagmiConfig.ts
imykolenko: thx, I missed that. NativeSymbol definition is already there, of course it makes sense to use decimals from loaded configuration.

PR #287, opened 2024-06-11
CHAIN-256: move DB migration and blockchain handlers to new "ring" co…

bholzman:
…ntainer

### DIFF
--- a/terraform/test/main.tf
+++ b/terraform/test/main.tf
@@ -28,6 +28,7 @@ module "api" {
   source            = "../modules/ecs_task"
   name_prefix       = local.name_prefix
   task_name         = "api"
+  include_command   = true
   image             = "backend"
   ecs_cluster_id    = module.ecs.cluster.id
   app_ecs_task_role = module.ecs.app_ecs_task_role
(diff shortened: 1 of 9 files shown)


terraform/test/main.tf
```diff
@@ -176,15 +193,18 @@ module "bastion" {
...
-  vpc             = module.vpc.vpc
-  aws_region      = var.aws_region
-  ci_role_arn     = data.terraform_remote_state.shared.outputs.ci_role_arn
+  source         = "../modules/rds"
+  name_prefix    = local.name_prefix
+  subnet_id_1    = module.vpc.private_subnet_id_1
+  subnet_id_2    = module.vpc.private_subnet_id_2
+  instance_class = "db.t3.medium"
+  security_groups = [
+    module.api.security_group_id, module.bastion.security_group.id, module.sequencer.security_group_id,
```
imykolenko, line 202: should ring's security group also be added here?

imykolenko (approved):
👍

terraform/test/main.tf
bholzman: yes, definitely, thanks!

PR #288, opened 2024-06-11
allow market orders using percent of limit

bflood366:


backend/src/main/kotlin/co/chainring/apps/api/model/Orders.kt
```diff
@@ -34,6 +35,7 @@ sealed class CreateOrderApiRequest {
         override val amount: BigIntegerJson,
         override val signature: EvmSignature,
         override val verifyingChainId: ChainId,
+        val percentage: Percentage? = null,
```
bholzman, line 38: would probably be cleaner to instead make `amount` a sealed class with `Fixed` or `Percentage` sub-classes. This has the benefit of being backwards compatible of course, but that doesn't matter yet.

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
```diff
@@ -82,9 +83,10 @@ sealed class EIP712Transaction {
         val sender: Address,
         val baseToken: Address,
         val quoteToken: Address,
-        val amount: BigIntegerJson,
+        val amount: BigIntegerJson?,
```
bholzman, line 86: this could then use that same sealed class

backend/src/main/kotlin/co/chainring/core/model/Percentage.kt
```diff
@@ -0,0 +1,15 @@
+package co.chainring.core.model
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+@JvmInline
+value class Percentage(val value: Int) {
+    init {
+        require(value in 0..MAX_VALUE) { "Invalid percentage" }
```
bholzman, line 9: fwiw, a percentage of 0 is probably not something we want to be possible

web-ui/src/components/Screens/HomeScreen/swap/SwapModal.tsx
```diff
@@ -180,9 +180,15 @@ export function SwapModal({
                       : sr.handleBaseAmountChange
                   }
                   sellAssetsNeeded={sr.sellAssetsNeeded}
+                  showMax={(sr.topBalance?.available ?? BigInt(0)) > BigInt(0)}
```
bholzman, line 183: I guess technically we ought to be looking at the limit here, not their available balance

bholzman (approved):
Looks good!

backend/src/main/kotlin/co/chainring/core/evm/EIP712Transaction.kt
bflood366: used same sealed class here

backend/src/main/kotlin/co/chainring/core/model/Percentage.kt
bflood366: change to 1 being min

web-ui/src/components/Screens/HomeScreen/swap/SwapModal.tsx
bflood366: changed to use the limit here

PR #290, opened 2024-06-12
CHAIN-310: fix flicker from spurious resubscribes

bholzman:


imykolenko (approved):
looks good!

web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
```diff
@@ -750,8 +759,8 @@ export function SwapInternals({
 
   const sellAssetsNeeded = useMemo(() => {
     return topSymbol.name === quoteSymbol.name
-      ? notional + fee - (quoteLimit || 0n)
-      : baseAmount - (baseLimit || 0n)
+      ? notional + fee - (quoteLimit ?? notional + fee)
+      : baseAmount - (baseLimit ?? baseAmount)
```
imykolenko, line 763: not fully in the context, this changes behavior to opposite. When quoteLimit is not defiled resulting value will be 0 instead of `notional + fee`

bholzman: yes, this was intentional -- the amount and the limit change asynchronously so when there's no limit defined, the idea is to set sellAssetsRequired to 0 so an insufficient balance message doesn't flash. probably not a problem after fixing the websocket resubscribe, but seems like preferable semantics anyway

imykolenko: thank for explaining!

PR #291, opened 2024-06-13
Make Telegram trading bot stateless

antonyshchenko:
- also implemented price estimation based on the current order book (CHAIN-305)
- will move bot code into backend package with a separate PR
- added `DepositStatus.SentToSequencer` and changed `BlockchainDepositHandler` and `SequencerResponseProcessorService` so that deposit is now completed by `SequencerResponseProcessorService` - this way deposit is marked as completed and balances are updated in the same transaction


sequencer/src/main/kotlin/co/chainring/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -65,25 +65,23 @@ object SequencerResponseProcessorService {
...
                 request.balanceBatch!!.withdrawalsList.forEach { withdrawal ->
-                    WithdrawalEntity.findById(withdrawal.externalGuid.withdrawalId())?.let { withdrawalEntity ->
-                        val balanceChange = response.balancesChangedList.firstOrNull { it.wallet == withdrawal.wallet }
-                        if (balanceChange == null) {
-                            withdrawalEntity.update(WithdrawalStatus.Failed, error(response, "Insufficient Balance"))
-                        } else {
-                            handleSequencerResponse(request = request, response = response, ordersBeingUpdated = listOf())
-                            withdrawalEntity.update(WithdrawalStatus.Sequenced, null, actualAmount = balanceChange.delta.toBigInteger().negate())
-                        }
+                    val withdrawalEntity = WithdrawalEntity.findById(withdrawal.externalGuid.withdrawalId())!!
```
antonyshchenko, line 68: Previously withdrawal would just get stuck forever in Pending state if created and sent to sequencer within a singe db transaction. It was working for the API since `ExchangeApiService` wraps only the `WithdrawalEntity` creation into `transaction` but was problematic for trading bot because it executes `ExchangeApiService.withdraw()` in `transaction`.

sequencer/src/main/kotlin/co/chainring/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -65,25 +65,23 @@ object SequencerResponseProcessorService {
...
+                    if (balanceChange == null) {
+                        withdrawalEntity.update(WithdrawalStatus.Failed, error(response, "Insufficient Balance"))
+                    } else {
+                        handleSequencerResponse(request = request, response = response, ordersBeingUpdated = listOf())
+                        withdrawalEntity.update(WithdrawalStatus.Sequenced, null, actualAmount = balanceChange.delta.toBigInteger().negate())
                     }
                 }
 
                 request.balanceBatch!!.depositsList.forEach { deposit ->
+                    val depositEntity = DepositEntity.findById(deposit.externalGuid.depositId())!!
```
antonyshchenko, line 79: Changed it here as well just in case since there might be a chance that deposit record is not yet committed to db and we try to find it here and if it not found we ignore the deposit!

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V44_AddSentToSequencerStatusToDeposit.kt
```diff
@@ -0,0 +1,48 @@
...
+    enum class V44_DepositStatus {
+        Pending,
+        Confirmed,
+        SentToSequencer,
+        Complete,
+        Failed,
+    }
+
+    object V44_DepositTable : GUIDTable<DepositId>("deposit", ::DepositId) {
+        val walletGuid = reference("wallet_guid", WalletTable).index()
```
bholzman, line 25: should define `V44_WalletTable` and `V44_SymbolTable` stubs and use those here

sequencer/src/main/kotlin/co/chainring/sequencer/apps/services/SequencerResponseProcessorService.kt
bholzman: if the deposit is not in the DB won't this throw a NPE?

bholzman (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V44_AddSentToSequencerStatusToDeposit.kt
antonyshchenko: yup, thanks

PR #293, opened 2024-06-13
Move telegram bot into backend

antonyshchenko:


terraform/demo/main.tf
```diff
@@ -28,6 +28,7 @@ module "api" {
   source            = "../modules/ecs_task"
   name_prefix       = local.name_prefix
   task_name         = "api"
+  include_command   = true
```
antonyshchenko, line 31: Not sure why this was missing here and in a few other places

bholzman: wow, oops... so I guess that means both ring container and the 2 api containers in demo have been running both the api and the ring apps! I guess this is probably the source of those weird "block height" errors we saw from demo

bholzman (approved):
Looks good!

PR #294, opened 2024-06-13
CHAIN-313: add missing indexes on chain_settlement_batch table

bholzman:


antonyshchenko (approved):
Nice! 👍

imykolenko (commented):
Thanks for taking care so quickly!

PR #298, opened 2024-06-15
CHAIN-315: handle wallet connected with invalid chain

bholzman:
https://github.com/Chainring-Inc/chainring-monorepo/assets/2530842/230ef080-5426-454e-9f95-eb15c3166cc2


imykolenko (approved):
looks great

PR #303, opened 2024-06-18
Telegram mini-app deployment

antonyshchenko:


.github/workflows/CI.yml
```diff
@@ -58,9 +62,20 @@ jobs:
...
       environment: test
 
+  deploy_telegram_mini_app:
+    name: Deploy Telegram Mini App
+    if: ${{ github.event_name == 'push' }}
+    needs: [ build_push_backend, build_telegram_mini_app ]
+    permissions:
+      id-token: write # needed for aws credentials
+      contents: read
+    uses: ./.github/workflows/deploy-web-ui.yml
```
imykolenko, line 72: It looks like should be `deploy-telegram-mini-app`

```suggestion
    uses: ./.github/workflows/deploy-telegram-mini-app.yml
```

terraform/demo/main.tf
```diff
@@ -230,6 +230,17 @@ module "web" {
   certificate_arn = data.aws_acm_certificate.chainring_us_east_1.arn
 }
 
+module "telegram_mini_app" {
+  source      = "../modules/telegram_mini_app"
+  name_prefix = local.name_prefix
+  zone        = data.terraform_remote_state.shared.outputs.zone
+  providers = {
+    aws.us_east_1 = aws.us_east_1
+  }
```
imykolenko, line 239: in `telegram_mini_app` there is no variable with name `providers`. What function it has?

.github/workflows/CI.yml
antonyshchenko: Yup, thanks

terraform/demo/main.tf
antonyshchenko: I'm not sure, but if I remove it Terraform would complain. I guess it has something to do with `required_providers` in `modules/telegram_mini_app/versions.tf`

PR #305, opened 2024-06-19
CHAIN-318,CHAIN-317: move Max button and fix limit price precision issue

bholzman:


sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
```diff
@@ -669,6 +674,7 @@ data class Market(
         } else if (order.type == Order.Type.MarketSell) {
             handleCrossingOrder(order)
         } else {
+            logger.debug { "Order ${order.guid}: Unknown order type ${order.type} rejected" }
```
imykolenko, line 677: it probably would make sense to log error in this case

imykolenko (approved):
👍

sequencer/src/main/kotlin/co/chainring/sequencer/core/Market.kt
bholzman: sure, will do

PR #306, opened 2024-06-19
CHAIN-321 apply d3 transitions only when page is visible

imykolenko:
Listening to `visibilitychange` event to determine if page is shown on the screen. Animated transitions are rendered only if page is shown to avoid building queue of pending animations.

### DIFF
--- a/web-ui/src/components/Screens/HomeScreen/OrderBookWidget.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/OrderBookWidget.tsx
@@ -17,6 +17,7 @@ import Decimal from 'decimal.js'
 import { useGesture } from '@use-gesture/react'
 import TradingSymbol from 'tradingSymbol'
 import SymbolIcon from 'components/common/SymbolIcon'
+import usePageVisibility from 'hooks/usePageVisibility'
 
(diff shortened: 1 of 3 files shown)


web-ui/src/components/Screens/HomeScreen/OrderBookWidget.tsx
```diff
@@ -377,12 +382,19 @@ function OrderBookChart({
...
       .select('.y-axis-order-book-last-price')
       .classed('hidden', !lastTrade.price)
-      // .transition()
-      // .duration(50)
-      .attr('transform', `translate(0,${yScale(parseFloat(lastTrade.price))})`)
+    if (renderAnimation) {
+      // @ts-expect-error Selection and Transition interfaces are compatible, casting only because of generic BaseType
+      lastPriceGroup = lastPriceGroup
+        .transition()
+        .duration(50) as typeof lastPriceGroup
```
imykolenko, line 392: selection and transition on top of selection have compatible interfaces. However I couldn't satisfy generics, therefore casting.

bholzman (approved):
Looks good, nice find!

PR #313, opened 2024-06-21
CHAIN-334 limit order price range

imykolenko:
- added `minAllowedBidPrice` and `maxAllowedOfferPrice` to market configuration. Actual values are set by sequencer on market creation
 - migration uses `last_price` to backfill data, values  would need to be adjusted after deployment 

==== Dashboard ====

Screenshot 2024-06-21 at 5 16 39 PM Screenshot 2024-06-21 at 5 16 47 PM Screenshot 2024-06-21 at 5 23 27 PM

==== Limit ====
Screenshot 2024-06-21 at 5 23 53 PM Screenshot 2024-06-21 at 5 24 05 PM Screenshot 2024-06-21 at 5 23 46 PM

### DIFF
--- a/sequencercommon/src/main/proto/sequencer.proto
+++ b/sequencercommon/src/main/proto/sequencer.proto
@@ -38,6 +38,8 @@ message MarketCreated {
   DecimalValue tickSize = 2;
   uint32 baseDecimals = 3;
   uint32 quoteDecimals = 4;
+  DecimalValue maxAllowedOffer = 5;
+  DecimalValue minAllowedBid = 6;
 }
 
 message FeeRates {
--- a/web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
@@ -58,6 +58,8 @@ export type SwapRender = {
   buyLimitPriceInputValue: string
   buyLimitPrice: Decimal
   limitPrice: Decimal
+  limitPriceTooLow: boolean
+  limitPriceTooHigh: boolean
   setPriceFromMarketPrice: (incrementDivisor?: bigint) => void
   noPriceFound: boolean
   canSubmit: boolean
@@ -260,7 +262,7 @@ export function SwapInternals({
       limitPrice
         .mul(new Decimal(10).pow(market.quoteSymbol.decimals))
         .floor()
-        .toString()
+        .toFixed(0)
     )
   }, [limitPrice, market])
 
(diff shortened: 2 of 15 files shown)


sequencercommon/src/main/proto/sequencer.proto
```diff
@@ -38,6 +38,8 @@ message MarketCreated {
   DecimalValue tickSize = 2;
   uint32 baseDecimals = 3;
   uint32 quoteDecimals = 4;
+  DecimalValue maxAllowedOffer = 5;
+  DecimalValue minAllowedBid = 6;
```
bholzman, line 42: will there be a problem parsing `MarketCreated` messages from before these were added if they are not optional?

web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
```diff
@@ -834,6 +843,25 @@ export function SwapInternals({
     return
   }, [orderBook, side, baseAmount, market, buyLimitPrice, quoteSymbol])
 
+  const [limitPriceTooLow, limitPriceTooHigh] = useMemo(() => {
+    const limitPriceTooLow =
+      isLimitOrder &&
+      limitPrice.gt(new Decimal(0)) &&
```
bholzman, line 849: is this actually needed? if it is less than zero it is definitely too low

bholzman (approved):

web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
imykolenko: Yes, it helps to reduce number of different inputs to be checked. So `limitPrice` is `0` until it input value is greater or at least equal to tick size. Limit widget without price specified would otherwise show `Price Too Low` message. 

Also Dashboard will show same on for example 0.04 on BTC/ETC market that stats from 0.05. In this way button just kept grayed out.

sequencercommon/src/main/proto/sequencer.proto
imykolenko: Good point. My thinking was that since the last checkpoint we did not create any new markets, therefore no problems expected. Of course it limits going back in history, not sure it is required though.

imykolenko: I already have sequencer queues locally from today, I will give it a try to be sure

imykolenko: Checked, old MarketCreated responses are parsed without problems into the new data structure as null. Generated java code is defensive it checks for nulls and returns `DecimalValue.getDefaultInstance()` in case of null.

Moreover, generated code with optional and non-optional max/min fields differs only in javadoc 😀 
```
     * <code>.sequencer.DecimalValue maxAllowedOffer = 5;</code>
```
vs
```
   * <code>optional .sequencer.DecimalValue maxAllowedOffer = 5;</code>
```

PR #320, opened 2024-06-25
Add ERC20 tokens support to faucet, allow only ERC20 tokens in testnet's faucet

antonyshchenko:


src/main/kotlin/co/chainring/tasks/SeedBlockchain.kt
```diff
@@ -55,35 +59,56 @@ fun seedBlockchain(fixtures: Fixtures): List<SymbolContractAddress> {
...
+        .withAlias("Waiting for transactions confirmations")
+        .pollInSameThread()
+        .pollDelay(Duration.ofMillis(100))
+        .pollInterval(Duration.ofMillis(100))
+        .atMost(Duration.ofMillis(10000L))
+        .until {
+            txHashes.all { (chainId, txHash) ->
+                blockchainClient(chainId).getTransactionReceipt(txHash)?.isStatusOK ?: false
+            }
+        }
```
antonyshchenko, line 110: Speeds up fixtures a little since we don't have to wait for each transaction to be completed sequentially

web-ui/.env.testnet
```diff
@@ -1 +1,3 @@
 ENV_API_URL=https://testnet-api.chainring.finance
+ENV_FAUCET_ENABLED=true
+ENV_NATIVE_FAUCET_ENABLED=false
```
bholzman, line 3: wondering if it would make more sense to include these flags in the backend config, so we don't have to keep the .env files in-sync with the env yamls?

bholzman (approved):
Looks great!

web-ui/.env.testnet
antonyshchenko: Good idea, I will add `faucetSupported` field to the symbol info returned from config endpoint

antonyshchenko: Done:
- added `faucetSupported` to `SymbolInfo` API model
- removed faucet-related vars from UI envs
- replaced the combination of `ENABLE_FAUCET_ROUTES` and `NATIVE_FAUCET_ENABLED` backend env vars with single `FAUCET_MODE` var

PR #322, opened 2024-06-26
CHAIN-343: prompt user to add certain tokens to their wallet

bholzman:


web-ui/src/utils/addSymbolsToWallet.ts
```diff
@@ -0,0 +1,26 @@
...
+  config.newSymbols?.map(async (sym) => {
+    if (
+      await window.ethereum?.request({
+        method: 'wallet_watchAsset',
+        params: {
+          type: 'ERC20',
+          options: {
+            address: sym.contractAddress,
+            // TODO: CHAIN-345 - improve cross-chain symbology
+            symbol: sym.name.replace(new RegExp('[0-9]+$', ''), ''),
```
bflood366, line 14: does a given ERC20 need to have the same address on all chains or does this just add for whatever the current chain id id. seems like the account configuration will return symbols for all chains if I understand correctly

bflood366 (approved):
looks great

web-ui/src/utils/addSymbolsToWallet.ts
bholzman: ah, actually there's an optional chainId you can add here, let me add that

bholzman: going to fix this in CHAIN-345 since the chain id is easily accessible there

PR #323, opened 2024-06-26
CHAIN-345: improve cross-chain symbology

bholzman:


backend/src/main/kotlin/co/chainring/core/model/db/Symbol.kt
```diff
@@ -118,6 +118,8 @@ class SymbolEntity(guid: EntityID<SymbolId>) : GUIDEntity<SymbolId>(guid) {
     var iconUrl by SymbolTable.iconUrl
     var addToWallets by SymbolTable.addToWallets
 
+    fun displayName() = this.name.replace(Regex(":.*"), "")
```
bflood366, line 121: there are a few places in `Output.kt` of the telegram bot that might need to use this method

bflood366 (approved):
looks great

backend/src/main/kotlin/co/chainring/core/model/db/Symbol.kt
bholzman: Yeah, I saw those but the chat interface doesn’t otherwise reference the chain, so we’d have to change things around more to avoid confusion about which “BTC” or “ETH” we’re talking about. Not sure that we will even need the chat interface at this point, so didn’t want to invest time on it here.

bflood366: 👍

PR #324, opened 2024-06-26
CHAIN-346: fix BTC/USDC product spec in demo

bholzman:


src/main/kotlin/co/chainring/tasks/fixtures/Fixtures.kt
```diff
@@ -82,8 +82,8 @@ fun getFixtures(chainringChainClients: List<BlockchainClient>) = Fixtures(
             Fixtures.Market(
                 baseSymbol = SymbolId(client.chainId, "BTC"),
                 quoteSymbol = SymbolId(client.chainId, "USDC"),
-                tickSize = "1.00".toBigDecimal(),
-                marketPrice = "68390.500".toBigDecimal()
+                tickSize = "25.00".toBigDecimal(),
+                marketPrice = "60812.500".toBigDecimal()
```
bflood366, line 86: The `order batches` test uses this market on both chains so it might need some adjustments

bflood366 (approved):
looks great

src/main/kotlin/co/chainring/tasks/fixtures/Fixtures.kt
bholzman: I believe that test was passing, but I'll confirm

PR #328, opened 2024-06-27
CHAIN-323 reaction time game

imykolenko:
https://github.com/Chainring-Inc/chainring-monorepo/assets/17568216/3301b77c-14c9-493c-97f4-edf36ba8bd97


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V49_TelegramMiniAppReactionTime.kt
```diff
@@ -0,0 +1,47 @@
...
+import co.chainring.core.model.telegram.miniapp.TelegramMiniAppUserId
+import co.chainring.core.model.telegram.miniapp.TelegramMiniAppUserRewardId
+import org.jetbrains.exposed.sql.SchemaUtils
+import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
+import org.jetbrains.exposed.sql.transactions.transaction
+
+@Suppress("ClassName")
+class V49_TelegramMiniAppReactionTime : Migration() {
+    object V49_TelegramMiniAppUserTable : GUIDTable<TelegramMiniAppUserId>("telegram_mini_app_user", ::TelegramMiniAppUserId) {
+        val gameTickets = long("game_tickets").default(5)
```
bholzman, line 17: maybe better to default to 0?

telegram-mini-app/src/components/MainScreen/game/ReactionGame.tsx
```diff
@@ -0,0 +1,351 @@
...
+const ticks = [
+  { description: 'A deep breath', label: '2.5s', timeMs: 2500 },
+  {
+    description: 'Light from the earth to the moon',
+    label: '1.3s',
+    timeMs: 1300
+  },
+  { description: 'Heartbeat', label: '500ms', timeMs: 500 },
+  { description: 'Blink of an eye', label: '150ms', timeMs: 150 },
+  { description: 'Lizard darts its tongue', label: '25ms', timeMs: 25 },
```
bholzman, line 250: nice!

telegram-mini-app/src/components/MainScreen/game/ReactionGame.tsx
```diff
@@ -0,0 +1,351 @@
...
+            }}
+          >
+            {tick.description}
+          </div>
+          {index === 0 && (
+            <div className="absolute top-8 h-0.5 w-[2000px] bg-white"></div>
+          )}
+        </div>
+      ))}
+      <div
```
bholzman, line 335: I think overall the animation is looking pretty good, but a few thoughts.
- I think we might want to play with the easing a bit here, I think the overall time it takes is probably good, but I think it might be easier to read if it slows down more in the middle of its path. Maybe we could implement the animation with css animations, which come with a bunch of different easing functions built-in?
- I think it would probably be better if the "You" arrow came in as if it were a tick itself instead of after

bholzman (approved):
Looking good!

backend/src/main/kotlin/co/chainring/apps/api/model/tma/ApiModels.kt
```diff
@@ -35,3 +37,15 @@ data class GetUserApiResponse(
...
 )
+
+@Serializable
+data class ReactionTimeApiRequest(
+    val reactionTimeMs: Long,
+)
+
+@Serializable
+data class ReactionsTimeApiResponse(
+    val percentile: BigDecimalJson,
```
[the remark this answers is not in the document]
imykolenko: Sure, it will. I thought we might want to display decimal places as well. Will change to int

telegram-mini-app/src/components/MainScreen/game/ReactionGame.tsx
imykolenko: Increased animation duration for 0.5s and used css transition type `ease-out`. It starts at a maximum speed and slows down until end, in this way mid way it's already slower and more readable.

imykolenko: Nice idea to animate 'You' arrow in the same way as other ticks. I guess it should go in between of other ticks. 
This opens up one more scenario, what if reaction time is greater that 2.5s? I can stop game displaying similar error to "Too early!". If scale is adjusted and 'You' is rendered first - it might look weird.

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V49_TelegramMiniAppReactionTime.kt
imykolenko: done

telegram-mini-app/src/components/MainScreen/game/ReactionGame.tsx
bholzman: I think that makes sense to stop the game, although 2.5 s could realistically be someone's reaction time... what if we just stop the game if the reaction time is greater than 5s and, in the case that the reaction time is between 2.5s and 5s, we could add another tick for 5s (e.g. stop a car from 175km/hr)?

imykolenko: - added extra tick for measured time > 2.5s
 - maximum time is set to 5s, then error message 'Try again' is shown and game restarts
 - changed trigger from `onClick` to `onTouchStart` and `onMouseDown` to get quicker response

PR #329, opened 2024-06-27
CHAIN-344: limit faucet to 1x per day for wallet / ip address

bholzman:


backend/src/main/kotlin/co/chainring/apps/api/FaucetRoutes.kt
```diff
@@ -54,33 +55,66 @@ class FaucetRoutes(private val faucetMode: FaucetMode, blockchainClients: Collec
             )
         } bindContract Method.POST to { request ->
             val payload = requestBody(request)
-            val symbol = transaction { SymbolEntity.forName(payload.symbol) }
+            val sourceAddress = request.source?.address
```
bholzman, line 58: I'm not actually sure if this will be the user's real IP address or not, so may need further adjustment if not

backend/src/main/kotlin/co/chainring/core/model/db/FaucetDrip.kt
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        fun eligible(symbol: SymbolEntity, walletAddress: Address, ipAddress: String): Boolean {
+            val oneDayAgo = Clock.System.now().minus(1.days)
+            return (
+                FaucetDripTable.select(FaucetDripTable.guid.count()).where {
+                    FaucetDripTable.symbolGuid.eq(symbol.guid).and(
+                        FaucetDripTable.createdAt.greater(oneDayAgo),
+                    ).and(
+                        FaucetDripTable.ipAddress.eq(ipAddress) or FaucetDripTable.walletAddress.eq(walletAddress.value),
```
bflood366, line 52: so if I grok, this it is limited either by IP or wallet address. So from the same IP address I cannot could not fund 2 different wallets in the same day or from 2 different IPs I could not fund the same wallet.  Could this be a limit in a demo env if say Andy was using more than wallet or needed more that 1 unit of something.

bflood366 (approved):
looks great

backend/src/main/kotlin/co/chainring/core/model/db/FaucetDrip.kt
bholzman: That’s right, that seems to be what most faucets do. If Andy needs more in demo we can just send it from one of the seeded addresses.

backend/src/main/kotlin/co/chainring/core/model/db/FaucetDrip.kt
```diff
@@ -0,0 +1,64 @@
...
+            return FaucetDripEntity.new(FaucetDripId.generate()) {
+                this.createdAt = Clock.System.now()
+                this.symbol = symbol
+                this.walletAddress = walletAddress.value
+                this.ipAddress = ipAddress
+            }
+        }
+
+        fun eligible(symbol: SymbolEntity, walletAddress: Address, ipAddress: String): Boolean {
+            val oneDayAgo = Clock.System.now().minus(1.days)
```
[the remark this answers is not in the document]
bholzman: Maybe, easy enough to add that if/when we need it

backend/src/main/kotlin/co/chainring/apps/api/FaucetRoutes.kt
imykolenko: I think it will be IP address of the ALB. Client's ip address should be available as `X-Forwarded-For` request header

bholzman: right, right, thanks, I'll adjust accordingly

backend/src/main/kotlin/co/chainring/core/model/db/FaucetDrip.kt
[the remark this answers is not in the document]
bholzman: I'll go ahead and make it configurable, might as well use a much smaller time limit for demo

PR #335, opened 2024-06-28
CHAIN-258: Provide mechanism to rollback a specific trade

bflood366:


integrationtests/src/test/resources/log4j2.xml
```diff
@@ -17,7 +17,7 @@
         <Logger name="co.chainring" level="debug" additivity="false">
             <AppenderRef ref="AsyncAppender"/>
         </Logger>
-        <Logger name="co.chainring" level="error" additivity="false">
+        <Logger name="co.chainring.core.repeater.tasks" level="error" additivity="false">
```
bflood366, line 20: I made this change since I was not getting logs while running integration tests locally - the GasMonitorTest still works after this change

contracts
bflood366, line 1: it looks like one of the game PR brought in an older version of contracts submodule - this just puts the correct one back in main. The generated code/stubs were ok.

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V53_AddPendingRollbackStatusToTrade.kt
```diff
@@ -0,0 +1,37 @@
...
+import co.chainring.core.db.updateEnum
+import co.chainring.core.model.db.GUIDTable
+import co.chainring.core.model.db.PGEnum
+import co.chainring.core.model.db.TradeId
+import org.jetbrains.exposed.sql.transactions.transaction
+
+@Suppress("ClassName")
+class V53_AddPendingRollbackStatusToTrade : Migration() {
+
+    enum class V52_SettlementStatus {
```
bholzman, line 13: nit: rename to `V53_SettlementStatus` and `V53_TradeTable` ?

integrationtests/src/test/resources/log4j2.xml
bholzman: weird, I don't understand that -- seems like the `<AppenderRef ref="AsyncAppender"/>` below ought to take care of that

bholzman (approved):
Looks good!

integrationtests/src/test/resources/log4j2.xml
bflood366: so I think the problem was there was 2 logger entries for the same package `co.chainring` - the first with level 'debug' which means debug and higher will be logged, but then the 2nd with level `error` so only error would be logged. That seems to take precedences for that package over the first one so I would only see error logs when running integration tests locally. If I reversed their order in log4j.xml then I would see the debug logs but then the GasMonitor test failed.

PR #339, opened 2024-07-01
CHAIN-323 optimize reaction-time game percentile calculation

imykolenko:
- do not record reaction for every user, instead use static table with 500 records (5000 ms / 10 ms step size). Good enough for percentile as 'int', response time is static
 - validate max reaction time on API endpoint, add more tests


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V50_TelegramMiniAppReactionTime.kt
```diff
@@ -2,17 +2,30 @@ package co.chainring.core.model.db.migrations
...
 import org.jetbrains.exposed.sql.SchemaUtils
 import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
 import org.jetbrains.exposed.sql.transactions.transaction
 
 @Suppress("ClassName")
 class V50_TelegramMiniAppReactionTime : Migration() {
+
+    @Serializable
+    @JvmInline
+    value class V51_TelegramMiniAppGameReactionTimeId(override val value: String) : EntityId {
```
bholzman, line 21: nit: this is V50

bholzman (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/model/db/migrations/V50_TelegramMiniAppReactionTime.kt
imykolenko: thx, updated!

PR #343, opened 2024-07-02
CHAIN-351 (2): some terraform/config tweaks for testnet; migration fix

bholzman:


bflood366 (approved):

terraform/modules/rds/main.tf
```diff
@@ -63,7 +63,7 @@ resource "aws_rds_cluster_instance" "db_instance" {
 
 resource "aws_iam_role" "rds_monitoring_role" {
   count = var.enable_advanced_monitoring ? 1 : 0
-  name  = "rds-monitoring-role"
+  name  = "${var.name_prefix}-rds-monitoring-role"
```
imykolenko, line 66: 👍

terraform/testnet/main.tf
```diff
@@ -79,7 +79,7 @@ module "rds" {
   instance_class = "db.r7g.xlarge"
   security_groups = [
     module.api.security_group_id, module.bastion.security_group.id, module.sequencer.security_group_id,
-    module.ring.security_group_id
+    module.ring.security_group_id, module.mocker.security_group_id
```
imykolenko, line 82: mocker does not connect to the database, this does not seem necessary

bholzman: ah, right, thanks, will remove

PR #347, opened 2024-07-02
CHAIN-164: add min fee for orders

bflood366:
If the quantity is too low then a `Quantity Too Low` error is shown on the Swap button. Same as how `Price Too Low` was handled. 

https://github.com/Chainring-Inc/chainring-monorepo/assets/162615268/f8cc29c0-4434-4df8-8061-b7fafb305f9d

### DIFF
--- a/sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
+++ b/sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
@@ -66,7 +66,13 @@ class SequencerApp(
                 val marketId = MarketId(market.marketId)
                 var error: SequencerError? = null
                 if (state.markets.containsKey(marketId)) {
-                    error = SequencerError.MarketExists
+                    val currentMarket = state.markets.getValue(marketId)
+                    if (market.tickSize.toBigDecimal() != currentMarket.tickSize ||
+                        market.baseDecimals != currentMarket.baseDecimals ||
+                        market.quoteDecimals != currentMarket.quoteDecimals
+                    ) {
+                        error = SequencerError.MarketExists
+                    }
                 } else {
                     val tickSize = market.tickSize.toBigDecimal()
                     val halfTick = tickSize.setScale(tickSize.scale() + 1) / BigDecimal.valueOf(2)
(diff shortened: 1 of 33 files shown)


sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
```diff
@@ -66,7 +66,15 @@ class SequencerApp(
...
                 if (state.markets.containsKey(marketId)) {
-                    error = SequencerError.MarketExists
+                    val currentMarket = state.markets.getValue(marketId)
+                    if (market.tickSize.toBigDecimal() != currentMarket.tickSize ||
+                        market.baseDecimals != currentMarket.baseDecimals ||
+                        market.quoteDecimals != currentMarket.quoteDecimals
+                    ) {
+                        error = SequencerError.MarketExists
+                    } else {
+                        currentMarket.minFee = market.minFee.toBigInteger()
```
bflood366, line 76: I made this change so we can call the sequencer AddMarkets again but have it just change the minFee as long as ticksize, base decimals and quote decimals are same as the created market.

antonyshchenko: This behaviour is a little counter-intuitive. Why not add a `SetMarketMinFee` call?

web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
```diff
@@ -70,6 +70,7 @@ export type SwapRender = {
   handleMaxBaseAmount: () => void
   handleMaxQuoteAmount: () => void
   topLimit: bigint | undefined
+  quantityTooLow: boolean
```
antonyshchenko, line 73: Wondering why you chose `quantity` term here and not `amount`?

antonyshchenko (approved):
Looks great!

sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
bflood366: added a `SetMarketMinFees` method

web-ui/src/components/Screens/HomeScreen/swap/SwapInternals.tsx
bflood366: no reason - I changed to `amount` to be more consistent with other variable names

PR #349, opened 2024-07-03
CHAIN-341: ensure withdrawals and settlements are sequenced properly …

bflood366:
…on chain.


backend/src/main/kotlin/co/chainring/core/model/db/Trade.kt
```diff
@@ -66,6 +68,13 @@ object TradeTable : GUIDTable<TradeId>("trade", ::TradeId) {
                 settlementStatus.inList(pendingTradeStatuses)
             },
         )
+        index(
+            customIndexName = "trade_response_sequence_pending_settlement_status",
+            columns = arrayOf(responseSequence, settlementStatus),
+            filterCondition = {
+                settlementStatus.eq(SettlementStatus.Pending)
+            },
+        )
```
imykolenko, line 77: wow, assuming it is b-tree, idea is to keep sorted responseSequences for pending trades so that postgres can access min value just by looking at index

imykolenko (approved):
looks good!

PR #351, opened 2024-07-04
Chain 349/periodically restart anvil

imykolenko:
Schedule anvil to be restarted once a week in demo env. Also restart other services.

### DIFF
--- a/.github/workflows/anvil-restart-by-cron.yml
+++ b/.github/workflows/anvil-restart-by-cron.yml
@@ -1,0 +1,37 @@
+name: Restart anvil by cron
+
+on:
+  schedule:
+    # Run every Sunday at 3:00 AM UTC (11:00 PM NY, 05:00 AM Berlin)
+    - cron: '0 3 * * 0'
+
+# make sure it does not interfere with other jobs in progress
+concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/anvil-restart-by-cron.yml
```diff
@@ -0,0 +1,37 @@
...
+          python-version: '3.10'
+          cache: 'pip'
+      - name: Setup AWS
+        uses: ./.github/actions/setup-aws
+      - name: Stop and start
+        run: |
+          cd scripts
+          test -d venv || python3 -m venv venv
+          . ./venv/bin/activate && pip install --upgrade pip && pip install -r requirements.txt
+          python3 ecs-deploy.py stop --env demo --services mocker,telegrambot,api,ring
```
antonyshchenko, line 34: Why do we need to restart these?

imykolenko: no, we don't. My thinking was that I could avoid errors in logs when for example ring tries to settle trades while mocker is still running.

antonyshchenko: good point 👍

antonyshchenko (approved):

PR #352, opened 2024-07-04
Telegram mini app bot protection

antonyshchenko:
As a form of bot detection, if they score below 25ms 2x in a row they get silently marked as a bot. They still appear to earn points but their times are not recorded and whoever invited them does not receive a %age of their points.


backend/src/main/kotlin/co/chainring/core/model/telegram/miniapp/TelegramMiniAppGameUserReactionTime.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
+
+object TelegramMiniAppGameUserReactionTimeTable : LongIdTable("telegram_mini_app_game_user_reaction_time") {
+    val createdAt = timestamp("created_at")
+    val userGuid = reference("user_guid", TelegramMiniAppUserTable).index()
+    val reactionTimeMs = long("reaction_time_ms")
+}
+
+class TelegramMiniAppGameUserReactionTimeEntity(guid: EntityID<Long>) : Entity<Long>(guid) {
```
imykolenko, line 17: There is no need to store every reaction time for every user who played the game, it can be a lot of data.

What if we use 2 fields on user table: 'count consecutive reactions less than 25 ms' and 'is_bot'. Actually for 2 consecutive reactions will be enough for `is_bot` field to have 3 states (no, maybe, yes)

imykolenko (approved):
looks great!

PR #353, opened 2024-07-04
Nullable deposit block number

antonyshchenko:
Set block number once deposit it confirmed, otherwise keep null as value


bholzman (approved):
Looks great!

PR #354, opened 2024-07-04
CHAIN-370 moker configuration to stabilize market price, fixed liquidity placement

imykolenko:
- `_MAKER_PRICE_STABILIZATION` market configuration flag to make mocker place liquidity at baseline price
 - `_MAKER_LIQUIDITY_PLACEMENT` market configuration flag to allow percentage of balance OR fixed amount liquidity


mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
```diff
@@ -214,11 +233,11 @@ class Maker(
     }
 
     private fun adjustQuotes(market: Market, curPrice: BigDecimal) {
-        logger.debug { "$id: adjusting quotes in market ${market.id}, current price: $curPrice" }
+        logger.debug { "$id: adjusting quotes in market ${market.id}, current price: $curPrice, price override: $marketPriceOverride" }
```
bholzman, line 236: May not make much difference, but it may be more “realistic” if the anchor price can deviate slightly from exactly 1

bholzman (approved):
Looks good!

mocker/src/main/kotlin/co/chainring/mocker/core/Maker.kt
imykolenko: Sure, I'll add slight deviation to the order book, good point.  
Market price deviates on any trade because exactly `1.0` level is left free and then it depends on amount of market trade.

imykolenko: I added price movement function to the Maker for market stabilization. 

I also adjust market price by fraction of the difference between current market price and desired price. In this way bigger market orders have temporary influence on liquidity placement, which will revert with the next smaller trades.

I will let it run on my machine for a while to see picture over time

imykolenko: Runs well. It is incompatible with automatic takers that are driven by own price function because balances are exhausted quickly. I assume this is fine, for testnet we are not going to have takers, do we?

PR #356, opened 2024-07-05
CHAIN-365 lazily create blockchain_nonce entry if it does not exist

imykolenko:


backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainTransactionHandler.kt
```diff
@@ -61,7 +61,7 @@ class BlockchainTransactionHandler(
         workerThread = thread(start = true, name = "batch-transaction-handler-$chainId", isDaemon = true) {
             logger.debug { "Batch Transaction handler thread starting" }
             dbTransaction {
-                BlockchainNonceEntity.clearNonce(blockchainClient.submitterAddress, chainId)
+                BlockchainNonceEntity.clear(blockchainClient.submitterAddress, chainId)
```
antonyshchenko, line 64: 👍 nice renaming

antonyshchenko (approved):

PR #357, opened 2024-07-05
Replace eth_newFilter with listening for blocks and eth_getLogs call for deposit detection

antonyshchenko:


bholzman (approved):
Looks great! Just curious if you considered subscribing to block updates instead of polling (not sure this would be better, as it could be that some rpc providers don’t support it for the same reasons they don’t all support eth filters)

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainClient.kt
```diff
@@ -258,6 +259,18 @@ open class BlockchainClient(val config: BlockchainClientConfig) {
         return web3j.ethBlockNumber().send().blockNumber
     }
 
+    fun getLogs(blockNumber: BigInteger, address: Address): EthLog =
+        web3j.ethGetLogs(
+            EthFilter(
+                DefaultBlockParameter.valueOf(blockNumber),
+                DefaultBlockParameter.valueOf(blockNumber),
+                address.value,
+            ),
```
bflood366, line 268: not sure if its supported but seems like you can specify topics here and the topic could be the 4 byte signature of the DEPOSIT_EVENT - this might be more efficient than pulling all the events back and filtering on client side if it worked

antonyshchenko:
> if you considered subscribing to block updates instead of polling

I thought that since we do this change because eth_filter is not properly supported by the testnet provider, polling might be more reliable.

antonyshchenko:
@imykolenko has rightly pointed out that we should handle forks in `BlockProcessor` since otherwise we might miss some deposits. I'm going to implement it within this PR. 

The idea is to store some minimal block info e.g `hash`, `height`, `parent hash` in the database so we could detect forks and do rollbacks. I won't implement sending a compensating tx to sequencer since we already wait for a configured number of on-chain confirmations before marking deposit as confirmed and sending it to sequencer.

I will also add some logic to fail deposits that are still in the pending state after configured number of confirmations (currently they would stay in pending state forever).

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainClient.kt
antonyshchenko: `web3j` does not seem to support it, leaving it as is for now

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+
+            @Suppress("UNCHECKED_CAST")
+            (getLogsResult.logs as List<EthLog.LogObject>).forEach { logResult ->
+                processLog(logResult.get())
+            }
+        } else {
+            logger.error { "It looks like we forked, looking for split point starting at $blockNumber, chainId=$chainId" }
+
+            val blocksToRollback = BlockEntity
+                .getRecentBlocksUpToNumber(blockNumber - BigInteger.valueOf(5), chainId)
```
bholzman, line 95: probably worth either making this look-back configurable, or driving it off of the number of blocks we require before considering a transaction finalized

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
```diff
@@ -0,0 +1,167 @@
...
+                }
+
+            val blockNumbersToRollback = blocksToRollback.map { it.number }
+
+            logger.debug { "Rolling back blocks ${blockNumbersToRollback.joinToString(", ")}, chainId=$chainId" }
+
+            if (DepositEntity.countConfirmedOrCompleted(blockNumbersToRollback, chainId) > 0) {
+                logger.error { "Failed to rollback due to finalized deposits in blocks to rollback, chainId=$chainId" }
+            } else {
+                DepositEntity.markAsFailedByBlockNumbers(blockNumbersToRollback, chainId, error = "Fork rollback")
```
bholzman, line 108: there's a unique index on the transaction hash in the `deposit` table -- so seems if we mark the deposit as failed and the user subsequently retries it, we'd get a conflict. not sure what the best solution here is, but maybe make the index ignore failed deposits?

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
antonyshchenko: good point, changed from 5 to `BLOCKCHAIN_DEPOSIT_HANDLER_NUM_CONFIRMATIONS + 1`

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
antonyshchenko: We rely on finding a single deposit entry by tx has in a few places. I've added a `can_be_resubmitted` column to the `deposit` table and I'm setting it to `true` in case we fail deposit because we could not obtain tx receipt before timeout. Then in `DepositEntity.createOpdate` I update deposit status and creation time in case it was failed and has this flag. Also added a test for this.

PR #359, opened 2024-07-09
Set trade creation timestamp to be the timestamp of sequencer response

antonyshchenko:


sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
```diff
@@ -570,15 +583,19 @@ class SequencerApp(
...
                             val request = SequencerRequest.parseFrom(bytes.toByteArray())
                             val response = processRequest(request, dc.index(), startTime)
 
                             if (strictReplayValidation && response.sequence <= lastSequenceNumberProcessedBeforeRestart) {
                                 // validate actual response matches expected while replaying requests
                                 loadResponseFromOutputQueue(response.sequence)?.let {
-                                    val expectedResponse = it.toBuilder().setProcessingTime(response.processingTime).build()
+                                    val expectedResponse = it.toBuilder()
+                                        .setProcessingTime(response.processingTime)
+                                        .setCreatedAt(response.createdAt)
```
bholzman, line 596: Fwiw, this would fail to match if we are replaying a transaction which was initially processed without the code to add the createdAt time, since the replayed response would have createdAt but the original one would not. Maybe this is ok, but we would have to either reset demo when deploying this or temporarily disable the replay validation in demo when this is deployed.

sequencercommon/src/main/proto/sequencer.proto
```diff
@@ -111,6 +111,7 @@ message SequencerResponse {
   repeated WithdrawalCreated withdrawalsCreated = 13;
   repeated WithdrawalFee withdrawalFeesSet = 14;
   repeated MarketMinFee marketMinFeesSet = 15;
+  uint64 createdAt = 16;
```
bholzman, line 114: does this need to be optional to be able to parse responses created before this field was added?

bholzman (approved):

sequencer/src/main/kotlin/co/chainring/sequencer/apps/SequencerApp.kt
antonyshchenko: Will it actually fail? `it` is the response from the output queue and we convert it to the expected response by adding `processingTime` and `createdAt` from the response we have just calculated. Or am I missing something here?

bholzman: ah, sorry, I was mistaken

sequencercommon/src/main/proto/sequencer.proto
antonyshchenko: Seems to be safe to leave it as is. Here is a test I did:

```
fun main() {
    val v1Bytes = sequencerResponse {
        this.guid = "guid"
        this.sequence = 123L
        this.processingTime = 456L
    }.toByteArray()

    SequencerResponseV2.parseFrom(v1Bytes).also {
        println(it.createdAt)
    }
}
```

`SequencerResponseV2` has `createdAt` field defined as non optional, `SequencerResponse` is the definition from before this PR (no `createdAt` field).

It prints `0` when I run it.

antonyshchenko: I guess a check in response processor could be added so that we don't create a trade with timestamp from 1970, but that's more of an edge case for the situation when right after deployment the response processor is processing a response that was created with prev version of sequencer...

antonyshchenko: added that check to `SequencerResponseProcessorService.handleSequencerResponse` just in case

PR #365, opened 2024-07-10
Pin block number for eth_call invocations while processing settlement

antonyshchenko:


backend/src/main/kotlin/co/chainring/apps/ring/BlockchainTransactionHandler.kt
```diff
@@ -512,13 +513,12 @@ class BlockchainTransactionHandler(
...
     private fun submitToBlockchain(tx: BlockchainTransactionEntity) {
         val submitterNonce = BlockchainNonceEntity.getOrCreateForUpdate(blockchainClient.submitterAddress, chainId)
-        val blockNumber = blockchainClient.getBlockNumber()
         try {
             val nonce = submitterNonce.nonce?.let { it + BigInteger.ONE } ?: getConsistentNonce(submitterNonce.key)
             logger.debug { "sending Tx with nonce $nonce" }
             val txHash = sendPendingTransaction(tx.transactionData, nonce)
             submitterNonce.nonce = nonce
-            tx.markAsSubmitted(txHash, blockNumber)
+            tx.markAsSubmitted(txHash, blockNumber = null, lastSeenBlock = blockchainClient.getBlockNumber())
```
antonyshchenko, line 521: Had to change this because otherwise the `blockNumber` column of transaction record in DB was always set to the block number we have seen during tx submission and not the actual block number from tx receipt.

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainClient.kt
```diff
@@ -210,18 +244,35 @@ open class BlockchainClient(val config: BlockchainClientConfig) {
...
-        return exchangeContract.balances(address.value, tokenAddress.value).sendAsync().await()
+    private fun <A> exchangeContractCall(block: DefaultBlockParam, f: Exchange.() -> RemoteFunctionCall<A>): RemoteFunctionCall<A> {
+        val contract = loadExchangeContract(exchangeContractAddress)
+        contract.setDefaultBlockParameter(block.toWeb3j())
+        val startTime = Clock.System.now()
+        while (true) {
+            try {
+                return f(contract)
+            } catch (e: ContractCallException) {
+                if (e.message?.contains("BlockOutOfRangeError") != true || Clock.System.now() - startTime >= config.maxRpcNodeEventualConsistencyTolerance) {
```
antonyshchenko, line 255: Ticket mentioned that we would get some “header not found” error, but in my testing with local `anvil` I was seeing responses like this:
```
{"jsonrpc":"2.0","id":2,"error":{"code":-32602,"message":"BlockOutOfRangeError: block height is 6375 but requested was 3500"}}
```

which are turned into `ContractCallException` by `web3j`

bholzman: hmm, ok, the "header not found" is what I got from bitlayer's testnet RPC, so I guess it varies by JSON-RPC node. We should handle that too, and also see what base's RPC node returns. We should probably also start a confluence doc with things to assess before adding a new chain, and add this to it.

bholzman (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/blockchain/BlockchainClient.kt
antonyshchenko: Indeed, I've just checked Bitlayer and Sepolia, they return `header not found` as you said. Updated the logic here to cover both cases.

PR #378, opened 2024-07-12
CHAIN-372: UI changes for Back to back orders

bflood366:


web-ui/src/markets.ts
```diff
@@ -79,4 +96,93 @@ export default class Markets {
...
+                }
+              }
+            }
+          })
+      })
+      console.log(
+        'Back to back markets are ',
+        Array.from(backToBackMarketsById.keys()),
+        Array.from(backToBackMarketsById.values())
+      )
```
imykolenko, line 182: 😁

bflood366: removed this

PR #380, opened 2024-07-15
Make sure all trades from single Sequencer response are settled together

antonyshchenko:


imykolenko (approved):

backend/src/main/kotlin/co/chainring/core/model/db/Trade.kt
```diff
@@ -121,9 +123,42 @@ class TradeEntity(guid: EntityID<TradeId>) : GUIDEntity<TradeId>(guid) {
                 ?.let { it[TradeTable.responseSequence.min()]?.toLong() }
         }
 
-        fun findPending(limit: Int = 100): List<TradeEntity> {
+        fun findPendingForNewSettlementBatch(limit: Int = 100): List<TradeEntity> {
+            val pendingTradesCountByRespSeq = TradeTable
```
bflood366, line 127: This will always doing 2 queries if there are trades to settle, this one and the one below. Could you do the normal query to get all trades for settlement up to the limit and then only in the case where the number of returned trades is the limit, do an additional query for all pending trades above the last trades sequence id whose response sequence is the same as the last trade. This way you only have to do the second query if the first query returns the limit. It does mean you will settle above the limit but using number of trades as a limit is somewhat arbitrary since settlement is doing netted balance changes.

bflood366 (approved):
looks great

backend/src/main/kotlin/co/chainring/core/model/db/Trade.kt
antonyshchenko: Good idea 👍  implemented

PR #381, opened 2024-07-15
CHAIN-387: fix settlement issue with partial fills on back to back orders

bflood366:


backend/src/main/kotlin/co/chainring/core/model/db/Balance.kt
```diff
@@ -87,7 +87,7 @@ class BalanceEntity(guid: EntityID<BalanceId>) : GUIDEntity<BalanceId>(guid) {
                         balanceId = startingBalanceEntity?.guid?.value ?: BalanceId.generate(),
                         balanceBefore = startingBalance,
                         balanceAfter = when (change) {
-                            is BalanceChange.Delta -> BigInteger.ZERO.max((startingBalance ?: BigInteger.ZERO) + change.amount)
+                            is BalanceChange.Delta -> (startingBalance ?: BigInteger.ZERO) + change.amount
```
bflood366, line 90: I made this change because the balance can go negative if we have to rollback a settlement and this was making the balance zero here as oppose to negative

backend/src/main/kotlin/co/chainring/core/model/db/Trade.kt
```diff
@@ -151,7 +151,7 @@ class TradeEntity(guid: EntityID<TradeId>) : GUIDEntity<TradeId>(guid) {
             return TradeEntity.find {
                 TradeTable.settlementBatchGuid.eq(settlementBatch.guid) and
                     TradeTable.settlementStatus.eq(SettlementStatus.FailedSettling)
-            }.orderBy(TradeTable.sequenceId to SortOrder.ASC).toList()
+            }.orderBy(TradeTable.sequenceId to SortOrder.DESC).toList()
```
bflood366, line 154: technically this change was not needed after I made the change to allow a balance to go negative in DB, but this change will fail trades in reverse order so the balance changes are rolled back in sequencer in the reverse order to which they occurred.

antonyshchenko (approved):
Looks great!

PR #382, opened 2024-07-16
CHAIN-388: serve symbol icon URLs from backend

bholzman:


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V63_SymbolIconUrls.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.jetbrains.exposed.sql.transactions.transaction
+
+@Suppress("ClassName")
+class V63_SymbolIconUrls : Migration() {
+
+    override fun run() {
+        transaction {
+            exec(
+                """
+                UPDATE symbol SET icon_url = '/src/assets/' || LOWER(SPLIT_PART(name, ':', 1)) ||'.svg'
```
antonyshchenko, line 13: I'm not sure this is going to work as is. I'm not sure `Vite` will include these files into the build with current configuration. I've checked the `demo-chainring-web` S3 and it does not contain those files. Even if it would, it seems that they would have a fingerprint suffix in the name.

bholzman: ah, guess you're right, although it does work locally. I'll post those icons to S3 and change the URLs

antonyshchenko: Yeah, that would be the right way to do it considering that we plan to add an admin UI

bholzman: ok, done

antonyshchenko (approved):

PR #383, opened 2024-07-16
CHAIN-389: API for market and symbol CRUD

bholzman:


backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+                    val baseSymbol = SymbolEntity.forName(payload.id.baseSymbol())
+                    val quoteSymbol = SymbolEntity.forName(payload.id.quoteSymbol())
+                    val market = MarketEntity.create(
+                        baseSymbol = baseSymbol,
+                        quoteSymbol = quoteSymbol,
+                        tickSize = payload.tickSize,
+                        minFee = payload.minFee,
+                        lastPrice = payload.lastPrice,
+                    )
+                    sequencerClient.createMarket(
```
bflood366, line 186: do we want to check `response.error` here

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+                        marketId = market.id.value.value,
+                        tickSize = market.tickSize,
+                        baseDecimals = market.baseSymbol.decimals.toInt(),
+                        quoteDecimals = market.quoteSymbol.decimals.toInt(),
+                        minFee = market.minFee.fromFundamentalUnits(market.quoteSymbol.decimals),
+                    )
+                    sequencerClient.setMarketMinFees(mapOf(market.id.value to market.minFee))
```
bflood366, line 193: this is technically not needed here since if the market was created in the prior step the min fee was passed in and this just sets it to same value - you really only need to call this when updating

bflood366 (approved):
looks great

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
bholzman: ah, ok, I wasn't quite sure because `seedSequencer` always does it. I'll remove it from here

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+        "admin/symbol" / symbolNamePathParam meta {
+            operationId = "patch-symbol"
+            summary = "Patch Symbol"
+            security = signedTokenSecurity.and(adminSecurity)
+            tags += listOf(Tag("admin"))
+            returning(Status.OK)
+        } bindContract Method.PATCH to { symbolName ->
+            { request ->
+                val payload = requestBody(request)
+                runBlocking {
```
antonyshchenko, line 143: Seems like `runBlocking` is not needed here

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+            returning(Status.OK, responseBody to listOf())
+        } bindContract Method.GET to { _ ->
+            Response(Status.OK).with(
+                responseBody of transaction {
+                    SymbolEntity.all().map {
+                        AdminSymbol(
+                            chainId = it.chainId.value,
+                            name = it.name,
+                            description = it.description,
+                            contractAddress = it.contractAddress,
```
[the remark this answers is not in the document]
bholzman: maybe ... could just do it in the front-end too (and can just read the decimals from the contract there)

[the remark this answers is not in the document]
bflood366: yeah that works too

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
bflood366: actually wouldn't you want to update the withdrawal fee in the sequencer here

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+                        chainId = payload.chainId,
+                        name = payload.name.replace(Regex(":.*$"), ""),
+                        contractAddress = payload.contractAddress,
+                        decimals = payload.decimals,
+                        addToWallets = payload.addToWallets,
+                        withdrawalFee = payload.withdrawalFee,
+                        description = payload.description,
+                        iconUrl = payload.iconUrl,
+                    )
+                }
```
bflood366, line 94: you should call sequencer here to set the withdrawal fee here too

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -0,0 +1,260 @@
...
+                ),
+            )
+            returning(Status.CREATED)
+        } bindContract Method.POST to { request ->
+            transaction {
+                runBlocking {
+                    val payload = requestBody(request)
+                    val baseSymbol = SymbolEntity.forName(payload.id.baseSymbol())
+                    val quoteSymbol = SymbolEntity.forName(payload.id.quoteSymbol())
+                    val market = MarketEntity.create(
```
bflood366, line 179: there is technically a potential race condition here in the we are creating the db entry and calling the sequencer within the database tx. There is some logic in the sequencer response process that processes the AddMarket response and looks up the market and there is a chance this database tx may not be submitted when the response processor runs and the market would not exist then

bholzman: Thanks, yeah, that was causing test failures, already fixed it.

backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
bholzman: Thanks, yup, added

PR #385, opened 2024-07-17
CHAIN-392: admin api for managing admins and fee rates

bholzman:


backend/src/main/kotlin/co/chainring/apps/api/AdminRoutes.kt
```diff
@@ -349,12 +352,109 @@ class AdminRoutes(
...
+            operationId = "set-fee-rates"
+            summary = "Set Fee Rates"
+            security = signedTokenSecurity.and(adminSecurity)
+            tags += listOf(Tag("admin"))
+            receiving(requestBody to FeeRates(FeeRate(100), FeeRate(200)))
+            returning(Status.CREATED)
+        } bindContract Method.POST to { request ->
+            val payload = requestBody(request)
+            runBlocking {
+                val(origFeeRates, feeRates) = transaction {
```
bflood366, line 425: I guess technically you don't need to persist fees here and rollback below. If the sequencer call is successful then the response processor would persist the updated fees - its just a little async to this call so it might not be updated when you return from here which might complicate the tests a little

bflood366 (approved):
looks great

PR #386, opened 2024-07-17
CHAIN-326: linked signer

bflood366:
- This is setup so that auth token validation or EIP712 signature verification for orders or withdrawals require no database look ups for linked signers.
- There is a LinkedSignerService which on start up reads the linked signers from the DB and caches them. It also receives postgres notifies from `BlockProcessor` when linked signers are added or removed on chain for a wallet.


backend/src/main/kotlin/co/chainring/core/services/LinkedSignerService.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    fun getLinkedSigner(walletAddress: Address, chainId: ChainId): Address? = linkedSignerByChain[chainId]?.get(walletAddress)
+
+    fun createOrUpdateWalletLinkedSigner(walletAddress: Address, chainId: ChainId, linkedSignerAddress: Address) {
+        val walletEntity = WalletEntity.getOrCreate(walletAddress)
+        if (linkedSignerAddress == Address.zero) {
+            WalletLinkedSignerEntity.findByWalletAndChain(walletEntity, chainId)?.delete()
+        } else {
+            WalletLinkedSignerEntity.createOrUpdate(
+                wallet = WalletEntity.getOrCreate(walletAddress),
```
bholzman, line 85: nit: you already have `walletEntity`

bholzman (approved):
Looks good!

PR #388, opened 2024-07-18
Rollup trades in UI

antonyshchenko:
Trades are grouped in UI by taker order id. This means that:
- taker would see a group of trades for each of their orders
- maker would see a group of trades for each of taker's order that has matched against one or many maker's orders

https://github.com/user-attachments/assets/074a1c16-a82c-435d-ba7b-9260c7e3e734


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V63_AddCounterOrderIdToOrderExecution.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import co.chainring.core.db.Migration
+import org.jetbrains.exposed.sql.transactions.transaction
+
+@Suppress("ClassName")
+class V63_AddCounterOrderIdToOrderExecution : Migration() {
+    override fun run() {
+        transaction {
+            exec("ALTER TABLE order_execution ADD COLUMN counter_order_guid CHARACTER VARYING(10485760)")
+            exec("UPDATE order_execution SET counter_order_guid = (SELECT oe.guid FROM order_execution oe WHERE oe.trade_guid = trade_guid AND oe.role != role)")
```
antonyshchenko, line 11: This might require demo reset since it has 3M `order_execution` records in the database

web-ui/src/utils/test.ts
```diff
@@ -41,7 +41,7 @@ describe('PricesUtils', () => {
     }
   }
 
-  it('should work as a no-op', () => {
+  it.skip('should work as a no-op', () => {
```
antonyshchenko, line 44: These tests were causing test worker crash with `out of memory` error, disabled them for now and created [CHAIN-395](https://chainringinc.atlassian.net/browse/CHAIN-395) to fix that.

backend/src/main/kotlin/co/chainring/core/model/db/OrderExecution.kt
```diff
@@ -110,7 +115,7 @@ class OrderExecutionEntity(guid: EntityID<ExecutionId>) : GUIDEntity<ExecutionId
                 .where {
                     OrderTable.walletGuid.eq(wallet.guid) and OrderExecutionTable.timestamp.less(beforeTimestamp)
                 }
-                .orderBy(Pair(OrderExecutionTable.timestamp, SortOrder.DESC))
+                .orderBy(Pair(TradeTable.responseSequence, SortOrder.DESC), Pair(TradeTable.sequenceId, SortOrder.ASC))
```
antonyshchenko, line 118: This method is used to fetch latest trades to send via Websocket as well as by API endpoint that lists trades. It has `limit` param and I'm not sure what would be the best way to implement limiting here since trades are now grouped by UI and it would be nice if there were no partially filled groups.

One idea that I have is to add a table that would model trades group and then we would fetch the groups first with a `limit` and then fetch order executions that belong to those groups (relation by `order_execution.group_id` column).

My initial idea that I've posted to Slack was to fetch orders with a limit first and then trades for those orders, but given how we group trades there might be a case for maker where they have a group that contains trades from multiple of maker's orders (taker order was matched against multiple orders of the given maker).

Any other ideas/feedback would be welcome.

antonyshchenko: After discussion in Slack agreed on a new dedicated method for fetching latest trades just for UI. It should fetch latest `limit` set of distinct sequencer response numbers and then load trades created within those responses. This way we would never return an incomplete group to UI.

web-ui/src/components/common/SymbolAndChain.tsx
```diff
@@ -1,18 +1,29 @@
...
 }
 
-export function SymbolAndChain({ symbol, noIcon }: Props) {
+export function SymbolAndChain({ symbol, noIcon, className, iconSize }: Props) {
   return (
-    <div className="flex text-sm font-normal">
+    <div
+      className={classNames(
+        'flex font-normal',
+        className ? className : 'text-sm'
```
bholzman, line 17: nit: could just put the `text-sm` above, and have just `className` here -- if there's something in `className` that changes the text size, it would override the `text-sm`, and `classNames()` ignores null or undefined values.

web-ui/src/components/common/SymbolAndChain.tsx
```diff
@@ -1,18 +1,29 @@
...
+      )}
+    >
       {!noIcon && (
         <SymbolIcon
           symbol={symbol}
-          className="mr-2 inline-block size-6 align-top"
+          className={classNames(
+            'mr-2 inline-block align-top',
+            iconSize ? `size-${iconSize}` : 'size-6'
```
bholzman, line 25: note that if whatever `size-${iconSize}` evaluates to doesn't appear somewhere else explicitly, tailwind will not include it in the compiled css unless it's in the `safelist` in `tailwind.config.mjs`

bholzman (approved):
Looks great!

web-ui/src/components/common/SymbolAndChain.tsx
antonyshchenko: thanks, changed as you suggest

web-ui/src/components/common/SymbolAndChain.tsx
antonyshchenko: Did not know that, thanks. Decided to leave `size-6` and removed `iconSize` prop from the component.

PR #389, opened 2024-07-18
CHAIN-396: allow chainId for auth token to be specified

bflood366:


antonyshchenko (approved):
Looks good. Wondering if maybe `switchChain` method would be useful in `ApiClient`

bflood366:
> Looks good. Wondering if maybe `switchChain` method would be useful in `ApiClient`

good idea - I added a `switchChain` to the `ApiClient`

PR #393, opened 2024-07-19
CHAIN-394: remove order update

bflood366:


backend/src/main/kotlin/co/chainring/core/model/db/migrations/V66_AddCounterOrderIdToOrderExecution.kt
```diff
@@ -1,14 +1,26 @@
...
+        val counterOrderGuid = reference(
+            "counter_order_guid",
+            V66_OrderTable,
+        ).index().nullable()
+    }
     override fun run() {
         transaction {
-            exec("ALTER TABLE order_execution ADD COLUMN counter_order_guid CHARACTER VARYING(10485760)")
-            exec("UPDATE order_execution SET counter_order_guid = (SELECT oe.guid FROM order_execution oe WHERE oe.trade_guid = trade_guid AND oe.role != role)")
+            SchemaUtils.createMissingTablesAndColumns(V66_OrderExecutionTable)
```
bflood366, line 22: @antonyshchenko I had to fix this migration since it failed locally for me after rebasing since I had existing order executions in my local db.  The original migration was not creating the counter_order_guid and a FK to order.guid or putting an index on it and the query to backfill it for some reason the subquery was returning null so the column was updating to null so it failed on my local when it tried to make it non nullable on line 24.  I fixed the query to backfill it and it as least worked locally for me now.

antonyshchenko: 👍 thanks for taking care of it

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
```diff
@@ -251,8 +241,8 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
     }
 
     companion object : EntityClass<OrderId, OrderEntity>(OrderTable) {
-        fun batchUpdate(market: MarketEntity, wallet: WalletEntity, createAssignments: List<CreateOrderAssignment>, updateAssignments: List<UpdateOrderAssignment>, backToBackMarket: MarketEntity? = null) {
-            if (createAssignments.isEmpty() && updateAssignments.isEmpty()) {
+        fun batchUpdate(market: MarketEntity, wallet: WalletEntity, createAssignments: List<CreateOrderAssignment>, backToBackMarket: MarketEntity? = null) {
```
antonyshchenko, line 244: `batchUpdate` is not accurate an accurate name anymore after removing order updates

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/co/chainring/core/model/db/Order.kt
bflood366: changed name to `batchCreate`

bholzman (approved):

PR #395, opened 2024-07-22
Put Sequencer checkpoints into queue so they can be replicated

antonyshchenko:


sequencer/src/main/kotlin/co/chainring/sequencer/core/SequencerState.kt
```diff
@@ -70,88 +88,56 @@ data class SequencerState(
...
+                            withdrawalFee {
+                                this.asset = it.key.value
+                                this.value = it.value.toIntegerValue()
+                            }.toByteArray()
+                        },
+                    )
+                    .write("balances")
+                    .bytes(getBalancesCheckpoint().toByteArray())
+                    .write("markets")
+                    .sequence(
```
bholzman, line 133: the previous implementation had the advantage of not needing to hold the entire checkpoint data for all markets in memory at once, but here we are. I imagine we will change how the checkpoints work at some point to do less allocation, so maybe it is ok for now, but something to keep in mind.

sequencer/src/main/kotlin/co/chainring/sequencer/core/SequencerState.kt
```diff
@@ -54,10 +50,32 @@ data class SequencerState(
...
+            )
+
+            wire.read("withdrawalFees").sequence(withdrawalFees) { map, v ->
+                while (v.hasNextSequenceItem()) {
+                    val fees = WithdrawalFee.parseFrom(v.bytes())
+                    map[Symbol(fees.asset)] = fees.value.toBigInteger()
+                }
+            }
+
+            measureNanoTime {
```
bholzman, line 77: even though the stuff before this "should" be fast, might be worth capturing its time anyway

bholzman (approved):
Looks great!

sequencer/src/main/kotlin/co/chainring/sequencer/core/SequencerState.kt
antonyshchenko: Yeah, I know. I think we could optimize this if we bypass Protobuf as serialization format for market checkpoints and write directly to the `wire`, but that is based on my assumption that `wire().write()` writes directly to the memory mapped (off-heap) file (queue). If that is the case then for each market we could stream it's checkpoint data to the queue without allocating an intermediate buffer that has to fit the entire market. I can explore this, but I propose to do it with a separate ticket.

sequencer/src/main/kotlin/co/chainring/sequencer/core/SequencerState.kt
antonyshchenko: Yup, done

sequencer/src/main/kotlin/co/chainring/sequencer/core/SequencerState.kt
antonyshchenko: Created [CHAIN-406](https://chainringinc.atlassian.net/browse/CHAIN-406) to address this

PR #399, opened 2024-07-26
Market trades broadacsting over webscocket

antonyshchenko:
- `Trades` subscription topic was renamed to `MyTrades`
- `Trades`, `TradesCreated` and `TradesUpdated` message types were renamed to `MyTrades`, `MyTradesCreated` and `MyTradesUpdated`
- `Orders` subscription topic was renamed to `MyOrders`
- `Orders`, `OrdersCreated` and `OrdersUpdated` message types were renamed to `MyOrders`, `MyOrdersCreated` and `MyOrdersUpdated`
- Added new `MarketTrades` subscription topic and `MarketTradesCreated` message type
- trade record in `MarketTradesCreated` message is serialised into array like this: `[<trade id>, <trade type as "Buy" or "Sell">, <amount>, <price>, <timestamp as milliseconds since epoch>]`

### DIFF
--- a/backend/src/main/kotlin/co/chainring/apps/api/model/websocket/OutgoingWSMessage.kt
+++ b/backend/src/main/kotlin/co/chainring/apps/api/model/websocket/OutgoingWSMessage.kt
@@ -2,16 +2,36 @@ package co.chainring.apps.api.model.websocket
 
 import co.chainring.apps.api.model.Balance
 import co.chainring.apps.api.model.BigDecimalJson
+import co.chainring.apps.api.model.BigDecimalSerializer
+import co.chainring.apps.api.model.BigIntegerJson
+import co.chainring.apps.api.model.BigIntegerSerializer
 import co.chainring.apps.api.model.MarketLimits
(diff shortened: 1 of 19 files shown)


imykolenko (approved):
👍

backend/src/main/kotlin/co/chainring/apps/api/model/websocket/OutgoingWSMessage.kt
```diff
@@ -79,26 +99,105 @@ data class OHLC(
...
+        constructor(takerOrderExecution: OrderExecutionEntity) :
+            this(
+                takerOrderExecution.tradeGuid.value,
+                takerOrderExecution.order.side,
+                takerOrderExecution.trade.amount,
+                takerOrderExecution.trade.price,
+                takerOrderExecution.timestamp,
+            )
+
+        object AsArraySerializer : KSerializer<Trade> {
```
imykolenko, line 136: I'm a bit concerned that backward compatibility requires more care that when using json, but saving amount of data transferred is definitely goof thing.

PR #400, opened 2024-07-29
API endpoint for getting last price for market

antonyshchenko:
Also removed `minAllowedBidPrice` and `maxAllowedOfferPrice` since Sequencer is more dynamic now.

@bholzman I would like to remove `market.lastPrice` from the config endpoint because it is dynamic response and the only place I've found that uses it is here: https://github.com/Chainring-Inc/chainring-monorepo/commit/17d93805223b230bebff0d71473a97676994fdc2#diff-2fdbccec7dc901990d4b7e717860d55823f329211a0bb7a6dd8bb16f1034e824R99-R105 

I don't quite understand why those particular numbers are hardcoded. Was it for made specifically for cross-chain BTC swaps?

### DIFF
--- a/backend/src/main/kotlin/co/chainring/apps/api/ApiApp.kt
+++ b/backend/src/main/kotlin/co/chainring/apps/api/ApiApp.kt
@@ -130,6 +130,7 @@ class ApiApp(config: ApiAppConfig = ApiAppConfig()) : BaseApp(config.dbConfig) {
                             withdrawalRoutes.createWithdrawal,
                             withdrawalRoutes.listWithdrawals,
                             LimitsRoutes.getLimits,
+                            PriceRoutes.getLastPriceForMarket,
                         )
 
                         if (enableTestRoutes) {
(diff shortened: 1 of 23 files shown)


bholzman:
The logic with last price is used to pick a logical default side for pairs - for example in a BTC/USDC market it’s very unlikely that somebody would want to see the price in terms of btc, it would almost always be in terms of usdc. So the logic is trying to pick a default side with the larger price.  For markets such as BTC/BTC or USDC/USDC or RING/RING there isn’t a natural default side so it uses the side based on if the user is buying or selling. The 1.1 and 0.9 are somewhat arbitrary but we’d need to pick some arbitrary threshold.

antonyshchenko:
@bholzman I see, going to leave the `market.lastPrice` in the config response as is then.

imykolenko (approved):
Looks good!

Agreed, min price probably does not make sense any longer, except we want to disallow someone to create orders to buy BTC for few USDCs. It is still covered by deposited amount, so why not.

Regarding max price, just occurred to me that in grpc, the levelIx, that represents number of ticks, is actually a int32 and looks can overflow. What if we instead keep it as a trading rule and also enforce check on API level?

antonyshchenko:
@imykolenko added a check, see https://github.com/Chainring-Inc/chainring-monorepo/pull/400/commits/d51b83c9d421ce4603dda7519dcaf2dc245d42c4

imykolenko (approved):

PR #407, opened 2024-07-30
Order book diff broadcasting over webscokets

antonyshchenko:


backend/src/main/kotlin/co/chainring/apps/api/model/websocket/OutgoingWSMessage.kt
```diff
@@ -54,29 +56,67 @@ sealed class Publishable
...
 }
 
 @Serializable
-data class LastTrade(
-    val price: String,
-    val direction: LastTradeDirection,
-)
+@SerialName("OrderBookDiff")
+data class OrderBookDiff(
+    val sequenceNumber: Long,
```
antonyshchenko, line 105: Sequence number is required for the client to figure out that it has missed an update and reconnect.

bholzman:
Saving a new snapshot of the entire order book on every order update seems like it will become a lot of data. Seems like there should be a better way to do this. The orders changed list coming from the sequencer should be enough information to construct an order book diff without needing to consult the last snapshot... so for the websocket API could we just calculate the snapshot on-demand when somebody first subscribes, and then just send them the diffs as they happen?

antonyshchenko:
>Saving a new snapshot of the entire order book on every order update seems like it will become a lot of data.

We are doing an insert or update, so an invariant of `0 <= snapshot records per market <= 1` is maintained all the time. I think this is more efficient than what we had before where we would re-calculate the order book and push it to Broadcaster every time we process a seq response with changed orders and also re-calculate it every time when someone wants to get it from the API. We could cache it in the API just as we do in the Broadcaster I guess, but here the cached value is essentially stored in the database, easy to read.

bholzman:
> > Saving a new snapshot of the entire order book on every order update seems like it will become a lot of data.
> 
> We are doing an insert or update, so an invariant of `0 <= snapshot records per market <= 1` is maintained all the time. I think this is more efficient than what we had before where we would re-calculate the order book and push it to Broadcaster every time we process a seq response with changed orders and also re-calculate it every time when someone wants to get it from the API. We could cache it in the API just as we do in the Broadcaster I guess, but here the cached value is essentially stored in the database, easy to read.

Ah, ok, I had missed that. Yes, makes sense, thanks.

bholzman (approved):
Looks great!

antonyshchenko:
But I agree, we could optimize it further and instead of re-calculating the entire order book in Sequencer response processor, recalculate only those levels that were affected by the changed orders. Would you like me to implement that change?

PR #414, opened 2024-08-01
CHAIN-418: deploy program to arch network

bflood366:


backend/src/main/kotlin/co/chainring/apps/ring/BitcoinContractsPublisher.kt
```diff
@@ -0,0 +1,52 @@
...
+                        deployContract(contractType, false)
+                    }
+                }
+            }
+        }
+    }
+
+    @OptIn(ExperimentalUnsignedTypes::class)
+    private fun deployContract(contractType: ContractType, firstTime: Boolean) {
+        if (firstTime) {
```
bholzman, line 34: maybe it would be better to use key-value store to keep track of if we've done the key exchange and generation instead of this contract flag?

build.gradle.kts
```diff
@@ -215,5 +215,19 @@ val compileContractsAndGenerateWrappers by tasks.register("compileContractsAndGe
...
+    exec {
+        isIgnoreExitValue = true
+        commandLine(
+            "bash", "-c", "cd ${projectDir}/contracts/arch && cargo build && cp $elfFile ${projectDir}/backend/src/main/resources"
+        )
+    }.also {
+        if (it.exitValue == 127) {
+            throw Exception("Cargo is missing, please follow the instructions: https://doc.rust-lang.org/cargo/getting-started/installation.html")
+        }
+    }
```
bholzman, line 232: add a "generated successfully" println here?

web-ui/src/components/Screens/HomeScreen.tsx
```diff
@@ -47,9 +47,9 @@ function HomeScreenContent() {
     marketsWithBackToBack
   } = useMemo(() => {
     const config = configQuery.data
-    const chainConfig = config?.chains.find(
-      (chain) => chain.id === (wallet.chainId || config.chains[0]?.id)
-    )
+    const chainConfig = config?.chains
+      .filter((chain) => chain.networkType === 'Evm')
```
bholzman, line 51: guess telegrambot may also need some filtering logic like this?

bholzman (approved):
Looks good!

build.gradle.kts
bflood366: done

web-ui/src/components/Screens/HomeScreen.tsx
bflood366: ah good point - added it

backend/src/main/kotlin/co/chainring/apps/ring/BitcoinContractsPublisher.kt
bflood366: changed to use KVstore

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
```diff
@@ -56,11 +57,15 @@ class BlockProcessor(
             while (true) {
                 try {
                     val blocksToProcess = transaction {
-                        (
-                            getLastProcessedBlock()?.number?.let { it + BigInteger.ONE }
-                                ?: DepositEntity.maxBlockNumber(chainId)?.let { it + BigInteger.ONE }
-                                ?: blockchainClient.getBlockNumber()
-                            ).rangeTo(blockchainClient.getBlockNumber())
+                        if (ChainEntity.findById(blockchainClient.chainId) != null) {
```
antonyshchenko, line 60: Not sure if changes here are really necessary since `BlockProcessor` is not created for `Bitcoin` networks from what I can see in this PR. For EVM chains `ChainWorker` does create it, but `BitcoinChainWorker` does not.

integrationtests/src/test/kotlin/co/chainring/integrationtests/api/DepositTest.kt
```diff
@@ -43,9 +44,11 @@ class DepositTest {
         wsClient.assertBalancesMessageReceived()
 
         val config = apiClient.getConfiguration()
-        assertEquals(config.chains.size, 2)
 
-        (0 until config.chains.size).forEach { index ->
+        val chains = config.chains.filter { it.networkType == NetworkType.Evm }
```
antonyshchenko, line 48: Noticed that this gets repeated in many tests. Might be worth adding `evmChains` property to `ConfigurationApiResponse`. It could look something like this:

```
    val evmChains: List<Chain>
        get() = chains.filter { it.networkType == NetworkType.Evm }
```

antonyshchenko (commented):
We should consider renaming `BlockchainClient` to `EvmClient` or something like that. Also `ChainManager` name is now too generic, should probably add an `EvmPrefix` there as well.

backend/src/main/kotlin/co/chainring/apps/ring/BlockProcessor.kt
bflood366: actually this is to fix an issue we see in the running tests in github - everything starts up with an empty db and block processor starts processing blocks but because the tests have not seeded the database yet with chains we get bombarded with errors since the block processor is trying to insert processed blocks (which have a FK to chains) but the chains are not populated.  Doesn't hurt anything just a bunch of noise in the logs that's sometimes confusing. This just prevents processing blocks for chains until they are in the DB.

bflood366:
> We should consider renaming `BlockchainClient` to `EvmClient` or something like that. Also `ChainManager` name is now too generic, should probably add an `EvmPrefix` there as well.

will add a ticket to for this.

PR #416, opened 2024-08-06
CHAIN-423: Rename to funkybit

bholzman:


backend/build.gradle.kts
```diff
@@ -99,7 +99,7 @@ java {
 }
 
 application {
-    mainClass.set("co.chainring.MainKt")
+    mainClass.set("xyzbit.MainKt")
```
antonyshchenko, line 102: Isn't this supposed to be `xyz.funkybit.MainKt` ?

README.md
```diff
@@ -1,6 +1,6 @@
-# ChainRing
+# funkybit
```
antonyshchenko, line 1: Is it `funkybit` or `Funkybit` ?

antonyshchenko (approved):

terraform/shared/variables.tf
```diff
@@ -11,27 +11,27 @@ variable "cidr_prefix" {
...
+  default = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIP8nY8Ix9r2Rh2Lxj0Kk9gQ0TupmCR4e+Sgh86sEPjUO loadtest-key@funkybit"
 }
 
 
 terraform {
   required_version = "1.5.7"
 
   backend "s3" {
-    bucket = "chainring-terraform-state"
+    bucket = "funkybit-terraform-state"
```
imykolenko, line 34: I guess this means completely new deployment, because this bucket is empty. Does it make sense to copy previous state to the new folder? Or maybe completely tear down old env using previous state to avoid non-tracked resources.

bholzman: yup, my plan was to tear down the old env

PR #417, opened 2024-08-06
Changes to enable telegram mini app in demo

bholzman:


bflood366 (approved):

telegram-mini-app/.env.demo
```diff
@@ -1,3 +1,3 @@
 ENV_API_URL=https://demo-api.chainring.co
-ENV_TG_BOT_ID=
-ENV_TG_APP_ID=
+ENV_TG_BOT_ID=ChainRingDevBot
```
antonyshchenko, line 2: Should probably be changed to `funkybit` as well

bholzman: can do that when we update the icons in the mini app

backend/src/main/kotlin/co/chainring/core/model/db/Balance.kt
```diff
@@ -118,51 +113,71 @@ class BalanceEntity(guid: EntityID<BalanceId>) : GUIDEntity<BalanceId>(guid) {
...
-                BatchUpdateStatement(BalanceTable).apply {
-                    updateAssignments.forEach { assignment ->
-                        addBatch(EntityID(assignment.balanceId, BalanceTable))
-                        this[BalanceTable.balance] = assignment.balanceAfter.toBigDecimal()
-                        this[BalanceTable.updatedAt] = now
-                        this[BalanceTable.updatedBy] = "system"
+                val (deltaUpdateAssignments, replaceUpdateAssignments) = updateAssignments.partition { it.changeType is BalanceChange.Delta }
+                if (deltaUpdateAssignments.isNotEmpty()) {
+                    TransactionManager.current().exec(
+                        """
```
antonyshchenko, line 146: For what it's worth, there is no need to fallback to the raw SQL query here. Exposed supports this in DSL: https://jetbrains.github.io/Exposed/deep-dive-into-dsl.html#update

backend/src/main/kotlin/co/chainring/core/model/db/Balance.kt
```diff
@@ -118,51 +113,71 @@ class BalanceEntity(guid: EntityID<BalanceId>) : GUIDEntity<BalanceId>(guid) {
...
+                    UPDATE ${BalanceTable.nameInDatabaseCase()}
+                       SET ${BalanceTable.balance.nameInDatabaseCase()} = ${BalanceTable.balance.nameInDatabaseCase()} + data_table.value
+                      FROM (
+                        SELECT unnest(?) as guid, unnest(?) as value
+                      ) AS data_table
```
antonyshchenko, line 151: Not sure what this part of the query does. Could you explain please?

backend/src/main/kotlin/co/chainring/core/model/db/Balance.kt
bholzman: That wouldn't work here since we are trying to update in a batch and we have a different specific delta to apply to each row.

We pass in an array of the balance guids to update and an array of the deltas to apply for the corresponding balance row. Then the unnest(?)s effectively turn those arrays into a table which we can join with do to the batch update.

antonyshchenko: Ah, nice, thanks for the explanation!

PR #418, opened 2024-08-06
Fix concurrent deposit withdrawal

bholzman:


bflood366 (approved):

antonyshchenko:
What exactly was happening here? Race condition during balance updates?

backend/src/main/kotlin/co/chainring/apps/ring/BlockchainTransactionHandler.kt
```diff
@@ -671,7 +672,7 @@ class BlockchainTransactionHandler(
     }
 
     private fun onWithdrawalComplete(withdrawalEntity: WithdrawalEntity, error: String?, withdrawAmount: BigInteger?) {
-        transaction {
+        transaction(Connection.TRANSACTION_SERIALIZABLE) {
```
antonyshchenko, line 675: Not sure if matters, but this is a nested transaction. Could be removed and `dbTransaction` calls in the worker thread's while loop could be updated to use `TRANSACTION_SERIALIZABLE` isolation level,

bholzman: Ah, that probably means this wasn't actually working, which makes sense since the failure seems to still be happening. What appears to be happening is that a deposit and withdrawal are updating the balances at roughly the same time. They are both doing delta updates to the balance, so if either transaction starts while the other one is still in progress, without serializable transactions, the ending balance in the DB would be wrong, since they both would see the initial balance before either update had happened.

antonyshchenko: `BlockchainDepositHandler` and `BlockchainTransactionHandler` are doing balance updates using deltas and those type of balance update is translated into `read current balance from db, calculate the new value using delta and replace the old value with the new value in db`. 

I wonder if something like  like `UPDATE balance SET balance = balance + delta` would be more reliable without changing the tx isolation level? My concern with serialization level change is  that it is this setting is located pretty far from the actual logic it affects and if we go with this approach we should at least add some comments to explain why we chose a non-default serialization level.

bholzman: I think you might be right -- I had thought about that but wasn't sure if the problem couldn't still happen, but experimenting a bit suggests that it will probably be ok.

PR #422, opened 2024-08-08
CHAIN-427: update with new logo

bholzman:


imykolenko (approved):
Looks good!

Curious how GH counts lines, it claims this PR removes 13k lines, but there is only one big file removed and it only accounts for 6k.

imykolenko (approved):
nevermind, there are 2.

PR #428, opened 2024-08-09
add cnames for funkybit and move corpweb over to new terraform state …

sburketx:
…bucket


terraform/corpweb/variables.tf
```diff
@@ -37,7 +37,7 @@ variable "static_ip_name" {
 
 variable "dns_zone" {
   type    = string
-  default = "funkybit.fun."
+  default = "chainring.co"
```
bholzman, line 40: What is this used for? Just wondering why it needs to stay chainring

bholzman (approved):
Looks good

terraform/corpweb/variables.tf
sburketx: we create the chainring.co A record for chainring from this terraform so if we change this to funkybit.fun, it takes the apex record away from chainring.co which breaks website resolution for those that still access the site using chainring. I suppose it may be better to look into taking away record creation entirely from the corpweb terraform and relegate it to the shared config with the other domain records.

bholzman: Got it, thanks, and yeah, that makes sense.

PR #431, opened 2024-08-13
Fix fork handling in BlockProcessor

antonyshchenko:
- `BlockProcessor` has an infinite loop where it calculates a sequence for block numbers to process based on the latest block stored in DB and latest block number returned from RPC node
- In that loop it would iterate over block numbers and for each block number it would fetch the block from RPC node and check if its parent hash equals to the hash of previous block in our DB, if yes - we would process the block and store it, otherwise that's a fork
- Fork handling was done incorrectly before this PR - we would rollback blocks until forking point and then continue block processing from the same place which is wrong. E.g: had a sequence of blocks 1..10 to process, on block 5 detected a fork, rolled back to block 3 and then continued processing from block 6 which obviously leads to another (false) fork detection alert and it continues forever.
- In this PR I've added the logic to break from the loop once we have handled the fork


bholzman (approved):
Looks good!

PR #436, opened 2024-08-14
Disable Bitcoin in integration tests workflow because it is not deployed yet

antonyshchenko:
`AppUnderTestRunner` thinks it is enabled and timeouts while waiting for Arch program to be deployed:

https://github.com/funkybit-labs/chainring-monorepo/blob/main/integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/testutils/AppUnderTestRunner.kt#L145-L157

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -42,6 +42,7 @@ jobs:
           EVM_NETWORK_URL_anvil: https://test-anvil.funkybit.fun
           EVM_NETWORK_URL_anvil2: https://test-anvil2.funkybit.fun
           INTEGRATION_RUN: 1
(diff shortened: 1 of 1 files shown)


bholzman (approved):

.github/workflows/integration-tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
           EVM_NETWORK_URL_anvil: https://test-anvil.funkybit.fun
           EVM_NETWORK_URL_anvil2: https://test-anvil2.funkybit.fun
           INTEGRATION_RUN: 1
+          BITCOIN_NETWORK_ENABLED: "false"
```
bflood366, line 45: ah my bad, thx for fixing

PR #438, opened 2024-08-14
fix for restoring when checkpoint is missing

bholzman:


sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/SequencerApp.kt
```diff
@@ -999,11 +999,24 @@ class SequencerApp(
...
-                    checkpointsQueue,
-                    expectedCycle = inputQueue.nextCycle(currentCycle, TailerDirection.BACKWARD),
-                )
+                var expectedCycle = inputQueue.nextCycle(currentCycle, TailerDirection.BACKWARD)
+                while (expectedCycle != inputQueue.firstCycle()) {
+                    logger.debug { "Restoring from checkpoint with expected cycle $expectedCycle" }
+                    try {
+                        state.load(
+                            checkpointsQueue,
+                            expectedCycle = inputQueue.nextCycle(currentCycle, TailerDirection.BACKWARD),
```
bflood366, line 1008: should this be `expectedCycle = expectedCycle`

bflood366 (approved):

sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/SequencerApp.kt
bholzman: yup thanks of course

sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/SequencerApp.kt
```diff
@@ -999,11 +1000,24 @@ class SequencerApp(
...
+                    logger.debug { "Restoring from checkpoint with expected cycle $expectedCycle" }
+                    try {
+                        state.load(
+                            checkpointsQueue,
+                            expectedCycle = expectedCycle
+                        )
+                        break
+                    } catch (e: Exception) {
+                        if (e.message?.contains("Invalid cycle in the checkpoint") == true) {
+                            expectedCycle -= 1
```
antonyshchenko, line 1014: I'm not 100% sure it is safe to just decrement by one here, but we could try and see if it works. Here is how `nextCycle()` is implemented: https://github.com/OpenHFT/Chronicle-Queue/blob/ea/src/main/java/net/openhft/chronicle/queue/impl/single/SingleChronicleQueue.java#L1192

bholzman: Ah, I guess because there could be gaps. Seems like that might not matter here since it simply won’t match anyway

bholzman: I changed it to 
```
                            expectedCycle = inputQueue.nextCycle(expectedCycle, TailerDirection.BACKWARD)
```

PR #445, opened 2024-08-15
Reduce tests flakiness

antonyshchenko:
- disabled repeater tasks automatic scheduling in tests so that they are run only when explicitly triggered from the test
- we now store gas alert status in DB and clear it between tests instead of just keeping it in memory since because app is not restarted in between individual tests, gas monitor alert could be triggered before it actually gets to the gas monitor test that expects no alerts in the beginning
- increased expected deposit confirmations from 1 to 2 to avoid race condition in `test on chain deposit detection - forks are handled`
- set Anvil mining interval to 60 minutes in test to make sure no mining is happening without explicit instruction to do so from the test (we had automining disabled previously, but that's not enough since that setting disables mining for each transaction, but does not disable interval mining)
- Changed `waitForFinalizedWithdrawal` test helper to explicitly expect certain withdrawal status instead of either `complete` or `failed` as this might lead to more determinism


bholzman (approved):
Looks good!

PR #448, opened 2024-08-15
state utxo onboarding (exchange and token state)

bflood366:


backend/src/main/kotlin/xyz/funkybit/core/model/BitcoinAddress.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import kotlinx.serialization.Serializable
+import org.bitcoinj.core.Address
+import org.bitcoinj.core.ECKey
+import org.bitcoinj.core.NetworkParameters
+import org.bitcoinj.script.Script
+
+@Serializable
+@JvmInline
+value class BitcoinAddress(val value: String) {
```
bholzman, line 11: I merged my EvmAddress/BitcoinAddress change so I think this can just get added there (in Address.kt). FYI, in my current branch to get bip322 signature verification working, I am extending BitcoinAddress to be its own sealed class with SegwitAddress, TaprootAddress, P2SHAddress, P2PKHAddress sub-classes. I'll change `fromKey` to return a `P2PKHAddress` after I rebase onto this

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchOnboardingTask.kt
```diff
@@ -0,0 +1,114 @@
...
+            val symbolsWithTokenState = ArchStateUtxoEntity.findAllTokenStateUtxo().map { it.symbolGuid }.toSet()
+            allSymbols.filterNot { symbolsWithTokenState.contains(it.guid) }
+        }
+    }
+
+    private fun initiateOnboarding(symbolEntity: SymbolEntity? = null): ArchStateUtxoEntity? {
+        val config = BitcoinClient.bitcoinConfig
+
+        val archNetworkAddress = ArchNetworkClient.getNetworkAddress()
+        val stateUtxoAmount = BigInteger("1500")
```
bholzman, line 68: thinking maybe we should ask them soon if they have plan for mainnet fees yet, as it could be more complicated than just a flat fee

backend/src/main/kotlin/xyz/funkybit/core/services/UtxoSelectionService.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    fun selectUtxos(address: BitcoinAddress, amount: BigInteger, fee: BigInteger): List<UnspentUtxo> {
+        return inputsSelector.selectInputs(
+            amount,
+            getUnspentUxtos(address).filter { it.reservedBy == null },
+            fee,
+        )
+    }
+
+    fun getUnspentUxtos(address: BitcoinAddress): List<UnspentUtxo> {
```
bholzman, line 28: nit: typo - Uxto -> Utxo

backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinUtils.kt
```diff
@@ -0,0 +1,133 @@
...
+import org.bitcoinj.core.TransactionOutput
+import org.bitcoinj.script.ScriptBuilder
+import org.bitcoinj.script.ScriptOpCodes
+import xyz.funkybit.core.blockchain.bitcoin.BitcoinClient
+import xyz.funkybit.core.model.BitcoinAddress
+import xyz.funkybit.core.model.db.UnspentUtxo
+import java.math.BigDecimal
+import java.math.BigInteger
+
+object BitcoinUtils {
```
bholzman, line 17: this is maybe more `ArchUtils`, no?

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        validateStateUtxo(
+            transaction {
+                waitFor {
+                    ArchStateUtxoEntity.findExchangeStateUtxo()?.status == StateUtxoStatus.Onboarded
+                }
+                ArchStateUtxoEntity.findExchangeStateUtxo()!!
+            },
+        )
+
+        // not verify token state utxo is onboaed
```
bholzman, line 50: nit: now

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+@ExtendWith(AppUnderTestRunner::class)
+class ArchOnboardingTest {
+    private val logger = KotlinLogging.logger {}
+
+    @Test
+    fun testOnboarding() {
+        Assumptions.assumeTrue((System.getenv("INTEGRATION_RUN") ?: "0") != "1")
+
+        transaction {
```
bholzman, line 30: should this use the `triggerRepeaterTaskAndWaitForCompletion` helper Anton added?

bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/services/UtxoSelectionService.kt
bflood366: oops - yep thx

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
bflood366: ok, will change

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
bflood366: yep

backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinUtils.kt
bflood366: yeah, will rename

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchOnboardingTask.kt
bflood366: This isn't a fee here really  - this is just the amount in the state utxo - but your point about the fees is well taken - For these onboarding state utxos I did not hard code the network fee - it is calculated based on tx vsize and current smart fee rate - I believe we will also need to do the same when calling the contract via arch RPC- the tx_hex is suppose to cover paying the bitcoin network fee for the tx that is put on chain after the contract call which includes the in/out state utxos and any other assets in/out utxos for deposits or withdrawals - so I think we need estimate that tx size and use the smart fee rate to calculate the fee - I will ask the Arch guys today for clarification around the fees

backend/src/main/kotlin/xyz/funkybit/core/model/BitcoinAddress.kt
bflood366: Why `P2PKHAddress` for the `fromKey` - I would think a P2WPKHAddress` would be better for our submitter since I believe it results in less fees.

PR #457, opened 2024-08-20
better fee calculations and allow unspent utxos in mempool to be avai…

bflood366:
…lable for utxo selection


antonyshchenko (approved):
Looks great!

PR #460, opened 2024-08-22
Deploy Bitcoin and Arch to test env

antonyshchenko:
Requires https://github.com/funkybit-labs/chainring-contracts/pull/23

### DIFF
--- a/scripts/envs/test.yml
+++ b/scripts/envs/test.yml
@@ -10,8 +10,11 @@ environment:
     JAVA_TOOL_OPTIONS: "-Xlog:gc+metaspace,gc+heap,gc:stdout:time,tags"
     ENABLE_TEST_ROUTES: "true"
     FAUCET_MODE: AllSymbols
-    BITCOIN_NETWORK_ENABLED: "false"
+    BITCOIN_NETWORK_ENABLED: "true"
+    BITCOIN_NETWORK_RPC_URL: "https://test-bitcoin.funkybit.fun"
+    BLOCK_EXPLORER_URL_BITCOIN: "https://test-mempool.funkybit.fun"
     REPEATER_AUTOMATIC_TASK_SCHEDULING: "false"
+    ARCH_NETWORK_RPC_URL: "https://test-arch.funkybit.fun"
   sequencer:
     QUEUE_HOME: "/data/queues/test"
     CHECKPOINTS_ENABLED: "false"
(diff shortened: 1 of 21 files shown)


scripts/envs/test.yml
```diff
@@ -36,27 +36,6 @@ environment:
   otterscan2:
     ERIGON_URL: https://test-anvil2.funkybit.fun
 services:
-  api:
```
antonyshchenko, line 39: Reordered services here to reflect dependencies. Also `ecs-deploy` would use the ordering from yaml file when optional `--services` argument is not provided.

scripts/envs/test.yml
```diff
@@ -74,6 +53,21 @@ services:
     image: 851725450525.dkr.ecr.us-east-2.amazonaws.com/otterscan
     count: 1
     environment: otterscan2
+  bitcoin:
+    count: 1
```
antonyshchenko, line 57: Image tag is static, so no need to specify it here. Also that ECS service contains multiple containers each using different image (bitcoin and fulcrum).

Same applies for `mempool` and `arch`.

Whenever we would need to upgrade `Arch` we can do it via `Terraform`.

terraform/demo/main.tf
```diff
@@ -108,8 +108,8 @@ module "anvil" {
   lb_dns_name                             = module.alb.dns_name
   zone                                    = data.terraform_remote_state.shared.outputs.zone
   tcp_ports                               = [8545]
-  health_check                            = "/"
-  health_check_status                     = "400"
+  health_check_http_path                  = "/"
```
antonyshchenko, line 111: Just some renaming

terraform/modules/ecs_task/main.tf
```diff
@@ -1,5 +1,6 @@
 locals {
-  http_api_port = var.allow_inbound ? var.tcp_ports[0] : 0
+  connect_to_load_balancer = var.lb_https_listener_arn != ""
+  inbound_port = var.allow_inbound ? var.tcp_ports[0] : 0
```
antonyshchenko, line 3: In the end I did not need this change, but decided to keep it since it provides more flexibility - for example we might want to allow inbound connections, but without connecting the ECS service to the loadbalancer.

terraform/modules/ecs_task/main.tf
```diff
@@ -106,41 +107,41 @@ resource "aws_security_group_rule" "security_group_ingress" {
...
+  from_port   = local.inbound_port
+  to_port     = local.inbound_port
   protocol    = "tcp"
   cidr_blocks = [var.vpc.cidr_block]
 }
 
 
-resource "aws_alb_target_group" "target_group" {
-  count                = var.allow_inbound ? 1 : 0
+resource "aws_lb_target_group" "target_group" {
```
antonyshchenko, line 117: `aws_alb_target_group` and `aws_lb_target_group` are the same

terraform/modules/ecs_task/main.tf
```diff
@@ -162,7 +163,7 @@ resource "aws_ecs_task_definition" "task" {
     merge(
       {
         name         = "${var.name_prefix}-${var.task_name}"
-        image        = "${local.account_id}.dkr.ecr.us-east-2.amazonaws.com/${var.image}:latest"
+        image        = var.image_is_external ? var.image : "${local.account_id}.dkr.ecr.us-east-2.amazonaws.com/${var.image}:latest"
```
antonyshchenko, line 166: Also ended up not needing this change, but decided to keep it as it provides more flexibility.

antonyshchenko:
Closing as we have decided against deploying Arch in test and demo env for now

PR #462, opened 2024-08-23
Optimize last trade price logic in order book snapshot calculation

antonyshchenko:


backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderBookSnapshot.kt
```diff
@@ -101,41 +102,60 @@ data class OrderBookSnapshot(
...
+
+                val prevTradePrice = if (latestOrderIds.size > 1) {
+                    val orderId = latestOrderIds[latestOrderIds.size - 2]
+                    weightedAvgPrice(
+                        latestTradesWithTakerOrdersForMarket
+                            .filter { it.second.guid.value == orderId }
+                            .map { it.first },
+                    )
+                } else {
+                    prevSnapshot?.last?.price ?: BigDecimal.ZERO
```
bholzman, line 138: if there's no prev snapshot, would lastTradePrice be a better default than zero?

bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderBookSnapshot.kt
antonyshchenko: In that case last trade direction would be `unchanged`. Is that what we want?

antonyshchenko: Actually, I've realised that we always have the prev snapshot, it would just be empty in case it is the first one.

bholzman: right, I thought it was just a first-time thing, in which case unchanged seems ideal

antonyshchenko: Made it non-nullable. The first-time snapshot would have `unchanged` last price direction.

PR #464, opened 2024-08-26
Represent fee rates as percentage in API

antonyshchenko:
- `0.01` in config API response means 1%
- We still use pips in admin UI, but I've added an info of how much 1 pip is in terms of percentages


HOW_TO_RESET_DEMO.md
```diff
@@ -118,8 +118,8 @@ export EVM_NETWORK_URL_Botanix="https://demo-anvil2.funkybit.fun"
 
 export SEQUENCER_HOST_NAME="localhost"
 export SEQUENCER_PORT="5338"
-export MAKER_FEE_RATE=0.01
-export TAKER_FEE_RATE=0.02
+export MAKER_FEE_RATE=0.0001
+export TAKER_FEE_RATE=0.0002
```
bholzman, line 122: actually, while we're changing these, we should go to 5 bps on the maker and 10 bps on the taker, so `0.0005` and `0.001`

scripts/envs/demo.yml
```diff
@@ -9,8 +9,8 @@ environment:
     SEQUENCER_PORT: "5337"
     JAVA_TOOL_OPTIONS: "-Xlog:gc+metaspace,gc+heap,gc:stdout:time,tags"
     SLACK_ERROR_REPORTING_CHANNEL_ID: C073ZA2VASW
-    MAKER_FEE_RATE: "0.01"
-    TAKER_FEE_RATE: "0.02"
+    MAKER_FEE_RATE: "0.0001"
```
bholzman, line 12: let's change to 5bps/10bps here and in testnet too

bholzman (approved):
Looks great!

scripts/envs/demo.yml
antonyshchenko: done

PR #465, opened 2024-08-27
Fix 'connection.connector.getProvider is not a function' error in UI when opening the page with connected EVM wallet

antonyshchenko:
During the initial load the `evmAccount.status` is in `reconnecting` state but `evmAccount.isConnected` is returning `true` 🤷 Because of this `signTypedData` was called before the connector is fully initialized and this resulted in a broken UI due to missing websocket connection.

Screenshot 2024-08-26 at 14 57 01


bholzman (approved):
Huh, weird, nice find

PR #466, opened 2024-08-27
CHAIN-468: in-progress

bholzman:


web-ui/src/auth.ts
```diff
@@ -1,10 +1,10 @@
 import { signTypedData } from '@wagmi/core'
 import { wagmiConfig } from 'wagmiConfig'
 import {
-  getBitcoinAccount,
-  getEvmAccount,
-  getPrimaryAddress,
-  getPrimaryCategory
+  getGlobalBitcoinAccount,
```
antonyshchenko, line 4: What does `global` mean here?

bholzman: since the auth stuff is outside of the react component hierarchy it can't use the bitcoin provider context, so this is storing the bitcoin account globally. originally I was just storing it in the window, but then I changed it to store it in local storage so the connection persists if you reload the tab

web-ui/src/auth.ts
```diff
@@ -69,14 +71,13 @@ async function signAuthToken(
             primaryType: 'Sign In'
           })
         : await (async () => {
-            const address = getBitcoinAccount()!.address
+            const bitcoinAccount = getGlobalBitcoinAccount()
+            console.log('global bitcoin account', bitcoinAccount)
```
antonyshchenko, line 75: Remove

antonyshchenko (approved):
Looks great!

PR #467, opened 2024-08-28
Feature/chain 452 identity linking

imykolenko:
- also show connect missing wallet category in swap widget
 - show primary wallet in header
 - on primary wallet disconnect authorized wallet becomes primary

=== EVM wallet authorize Bitcoin wallet ===

https://github.com/user-attachments/assets/72cf3edd-8631-4efb-a292-b699aa91f64b

=== Bitcoin wallet authorize EVM wallet ===

https://github.com/user-attachments/assets/47f0d3fa-d1db-4235-a09d-11175c3f8788


backend/src/main/kotlin/xyz/funkybit/core/model/telegram/bot/TelegramBotUser.kt
```diff
@@ -115,7 +115,8 @@ class TelegramBotUserEntity(guid: EntityID<TelegramBotUserId>) : GUIDEntity<Tele
 
     fun addWallet(privateKey: String): TelegramBotUserWalletEntity {
         return TelegramBotUserWalletEntity.create(
-            WalletEntity.getOrCreate(EvmAddress.fromPrivateKey(privateKey)),
+            // TODO do we really need to create a user here?
```
bholzman, line 118: probably worth it, seems like if we want the telegram bot to support BTC (or some other family of wallet like TON or SOL) we will need it

backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinSignatureVerification.kt
```diff
@@ -51,7 +51,7 @@ object BitcoinSignatureVerification {
             val pubkey =
                 ECKey.recoverFromSignature(recId, ECKey.ECDSASignature(r, s), Sha256Hash.of(msgHash), compressed)
             logger.debug { "Recovered pubkey $pubkey from ECDSA signature" }
-            return pubkey != null
+            return pubkey != null // TODO verify that recovered address matches the address provided in the message
```
bholzman, line 54: 👍

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/ApiClient.kt
```diff
@@ -72,12 +77,49 @@ data class ApiCallFailure(
...
+            }
+            is Bitcoin -> {
+                ecKey.signMessage(signInMessage.message + "\nAddress: ${signInMessage.address}, Timestamp: ${signInMessage.timestamp}")
+            }
+        }
+
+    fun address(): Address =
+        when (this) {
+            is EVM -> Address.auto(Credentials.create(ecKeyPair).address)
+            is Bitcoin -> BitcoinAddress.fromKey(NetworkParameters.fromID(NetworkParameters.ID_REGTEST)!!, ecKey)
```
bholzman, line 109: I suppose this makes sense since the integration tests should always be using regtest, but could also get the params from `BitcoinClient.getParams`

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/TestApiClient.kt
```diff
@@ -189,10 +193,30 @@ class TestApiClient(ecKeyPair: ECKeyPair = Keys.createEcKeyPair(), traceRecorder
     override fun markSymbolAsAdded(symbolName: String) =
         tryMarkSymbolAsAdded(symbolName).assertSuccess()
 
+    override fun linkWallets(bitcoinLinkAddressProof: BitcoinLinkAddressProof, evmLinkAddressProof: EvmLinkAddressProof) =
+        tryLinkWallets(LinkWalletsApiRequest(bitcoinLinkAddressProof, evmLinkAddressProof)).assertSuccess()
+
+    fun linkBitcoinWallet(bitcoinKeyPair: WalletKeyPair.Bitcoin) {
+        val bitcoinAddress = BitcoinAddress.fromKey(NetworkParameters.fromID(NetworkParameters.ID_REGTEST)!!, bitcoinKeyPair.ecKey)
```
bholzman, line 200: I guess same suggestion here

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/Wallet.kt
```diff
@@ -243,13 +237,13 @@ class Wallet(
             Symbol(symbol),
             amount,
             nonce,
-            blockchainClientsByChainId.getValue(currentChainId).signData(EIP712Helper.computeHash(tx, this.currentChainId, exchangeContractAddressByChainId.getValue(currentChainId)), linkedSignerEcKeyPair),
+            blockchainClientsByChainId.getValue(currentChainId).signData(EIP712Helper.computeHash(tx, this.currentChainId, exchangeContractAddressByChainId.getValue(currentChainId)), linkedSignerKeyPair?.let { it as WalletKeyPair.EVM }?.ecKeyPair),
```
bholzman, line 240: perhaps `(linkedSignerKeyPair as? WalletKeyPair.EVM)?.ecKeyPair` ?

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/api/WalletLinkingTest.kt
```diff
@@ -0,0 +1,424 @@
...
+import java.math.BigDecimal
+import kotlin.test.Test
+import kotlin.test.assertTrue
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+@ExtendWith(AppUnderTestRunner::class)
+class WalletLinkingTest : OrderBaseTest() {
+
+    @org.junit.jupiter.api.Test
```
bholzman, line 52: is there a reason not to import `Test`?

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/api/WalletLinkingTest.kt
```diff
@@ -0,0 +1,424 @@
...
+    fun `evm wallet link bitcoin wallet`() {
+        val evmWalletKeyPair = WalletKeyPair.EVM.generate()
+        val evmKeyApiClient = TestApiClient(evmWalletKeyPair)
+        val evmAddress = evmKeyApiClient.address as EvmAddress
+        assertTrue { evmKeyApiClient.getAccountConfiguration().linkedAddresses.isEmpty() }
+
+        val bitcoinKey = ECKey()
+        val bitcoinKeyApiClient = TestApiClient(WalletKeyPair.Bitcoin(bitcoinKey), chainId = ChainId(0u))
+        val bitcoinAddress = BitcoinAddress.fromKey(NetworkParameters.fromID(NetworkParameters.ID_REGTEST)!!, bitcoinKey)
```
bholzman, line 61: maybe DRY these 3 lines? or maybe we just wait for `CHAIN-484`?

backend/src/main/kotlin/xyz/funkybit/apps/api/WalletRoutes.kt
```diff
@@ -0,0 +1,216 @@
...
+import xyz.funkybit.core.model.EvmSignature
+import xyz.funkybit.core.model.db.ChainId
+import xyz.funkybit.core.model.db.WalletEntity
+import xyz.funkybit.core.model.db.WalletLinkProofEntity
+import xyz.funkybit.core.utils.bitcoin.BitcoinSignatureVerification
+import java.math.BigInteger
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+const val BASE_LINK_MESSAGE = "[funkybit] Please sign this message to link your wallets. This action will not cost any gas fees."
```
bholzman, line 41: I'm wondering if "link your wallets" is the best way we could express what's happening to the user. It's maybe not intuitive what they are agreeing to here. And as I think about it, what they're doing is not even really symmetric between their two wallets. When they sign-in with their first wallet, they are signing a message to prove that they control their wallet. Then, when they want to add a second wallet, they need to do two things: 1) prove they control the new wallet, and 2) prove that their original wallet authorizes this new wallet. And thinking ahead to when we support an additional wallet family (like say, TON or SOL), we would need 2) for just one of their existing wallets. So seems like maybe for a user who starts with EVM wallet A to add Bitcoin wallet B, they should sign a standard bitcoin sign-in message with wallet B and then sign a message like `[funkybit] Please sign this message to authorize Bitcoin wallet ${address}`. This action will not cost any gas fees.`. wdyt?

bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/Order.kt
```diff
@@ -13,14 +13,14 @@ import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
...
 import org.jetbrains.exposed.sql.batchInsert
+import org.jetbrains.exposed.sql.innerJoin
 import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
 import xyz.funkybit.apps.api.model.Order
 import xyz.funkybit.core.model.EvmSignature
 import xyz.funkybit.core.model.SequencerOrderId
 import xyz.funkybit.core.model.db.OrderExecutionTable.nullable
-import xyz.funkybit.core.model.db.migrations.V60_AddResponseSequenceToTradeAndWithdrawal.V60_TradeTable.default
```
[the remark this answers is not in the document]
imykolenko: yep 🤷‍♂️

backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinSignatureVerification.kt
imykolenko: implemented in https://github.com/funkybit-labs/chainring-monorepo/pull/469

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/TestApiClient.kt
imykolenko: makes sense!

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/api/WalletLinkingTest.kt
imykolenko: that's was auto-import. Addressed in https://github.com/funkybit-labs/chainring-monorepo/pull/469

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/Wallet.kt
imykolenko: nice, thx

backend/src/main/kotlin/xyz/funkybit/core/model/db/Wallet.kt
```diff
@@ -23,26 +25,58 @@ value class WalletId(override val value: String) : EntityId {
     override fun toString(): String = value
 }
 
+enum class WalletType {
```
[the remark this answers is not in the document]
imykolenko: was not sure about `type` either. Thanks for the suggestion, I like it.

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/api/WalletLinkingTest.kt
imykolenko: Sure, I introduces few helper methods to make it cleaner. Bigger changes will leave to CHAIN-484

backend/src/main/kotlin/xyz/funkybit/core/model/telegram/bot/TelegramBotUser.kt
imykolenko: This was a quick fix, for adding wallet from different families just need to lookup user.

However same telegram user can add multiple wallets from the same wallet family. Currently adding second same family wallet will fail as per design (unique constraint). Looks like it is a bigger topic how to support that, relaxing one-wallet-per-family constraint could be challenging

bholzman: Sorry, I wasn’t saying we need to support more than one wallet per family, just that we would need to support multiple families per telegram user

backend/src/main/kotlin/xyz/funkybit/apps/api/WalletRoutes.kt
imykolenko: yeah, message indeed may not provide full context of action. 
When we discussed UX I think we agreed that both wallets should allow user to see same data and perform same actions (except signing deposits and withdrawals), no matter which one was added first. And we should not require 'A' wallet to be always connected. In this context are wallets really not symmetric? 

In order to reduce invariants for the proof checks wallet links are modeled symmetric. Both wallets are cryptographically tight together by signing same (well, structurally different string vs EIP712) message containing own address, other wallet's address and timestamp. Both signed messages delivered to BE and accepted only when both are provided at the same time by one of the wallets. From the on-chain perspective verification check is same despite of order.
And yes, further ahead while adding 3rd wallet family we are likely going to ask only one of existing wallets to authorize, e.g. 'A' authorizes 'B', and later 'B' authorizes 'C'. This produces transitive 'A' to 'C' authorization from the perspective of proofs check. 

Regular 'B' wallet's sign-in message does not provide 'B' to 'A' authorization. Did you mean address 'A' should be  included? Also user in the UI already in the signed-in state, second sign-in might be misleading. 

I like your message proposal `[funkybit] Please sign this message to authorize Bitcoin wallet ${address}`.
What if we show it slightly modified for both wallets to suggest that it is mutual authorization?
EVM: `[funkybit] Please sign this message to authorize Bitcoin wallet ${address}`
Bitcoin: `[funkybit] Please sign this message to authorize EVM wallet ${address}`

bholzman: Yes, I remember we said that, but I'm questioning if that's actually important. We already know that this user controls wallet A. By giving us a "sign-in" message on wallet B, we know they control wallet B too. And by having wallet A "authorize" wallet B (and presenting both the sign-in message and the authorization message at the same time to the endpoint) we know that they are the same user. I'm not sure what we really add by making wallet B authorize wallet A too.

imykolenko: Thanks for the discussion during standup. I updated protocol to be as discribed in your comment:
A signs authorization message for B.
B signs regular sign-in message and calls BE with own auth token and signed authorization from A.

backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V77_User.kt
```diff
@@ -0,0 +1,98 @@
...
+            exec("CREATE TYPE WalletFamily AS ENUM (${enumDeclaration<V77_WalletFamily>()})")
+            SchemaUtils.createMissingTablesAndColumns(V77_WalletTable)
+            V77_WalletTable.selectAll().forEach { walletRecord ->
+                val userId = UserId.generate()
+                V77_UserTable.insert { userRecord ->
+                    userRecord[V77_UserTable.guid] = userId
+                    userRecord[V77_UserTable.createdAt] = Clock.System.now()
+                    userRecord[V77_UserTable.createdBy] = "V77_WalletTable"
+                }
+                walletRecord[V77_WalletTable.userGuid] = userId
```
bholzman, line 80: I needed this diff to get this migration to run:

```
diff --git a/backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V77_User.kt b/backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V77_User.kt
index 2af986b1..98d329e4 100644
--- a/backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V77_User.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V77_User.kt
@@ -9,6 +9,7 @@ import org.jetbrains.exposed.sql.insert
 import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
 import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.update
 import xyz.funkybit.core.db.Migration
 import xyz.funkybit.core.model.db.GUIDTable
 import xyz.funkybit.core.model.db.PGEnum
@@ -77,8 +78,10 @@ class V77_User : Migration() {
                     userRecord[V77_UserTable.createdAt] = Clock.System.now()
                     userRecord[V77_UserTable.createdBy] = "V77_WalletTable"
                 }
-                walletRecord[V77_WalletTable.userGuid] = userId
-                walletRecord[V77_WalletTable.walletFamily] = V77_WalletFamily.Evm
+                V77_WalletTable.update({V77_WalletTable.guid eq walletRecord[V77_WalletTable.guid]}) {
+                    it[userGuid] = userId
+                    it[walletFamily] = V77_WalletFamily.Evm
+                }
             }
             exec("ALTER TABLE wallet ALTER COLUMN user_guid SET NOT NULL")
             exec("ALTER TABLE wallet ALTER COLUMN wallet_family SET NOT NULL")
@@ -89,7 +92,7 @@ class V77_User : Migration() {
             // change wallet_guid in limit table to user_guid
             exec("ALTER TABLE \"limit\" DROP CONSTRAINT unique_limit")
             SchemaUtils.createMissingTablesAndColumns(V77_LimitTable)
-            exec("UPDATE \"limit\" SET user_guid = (SELECT user_guid FROM wallet WHERE wallet.user_guid = \"limit\".user_guid)")
+            exec("UPDATE \"limit\" SET user_guid = (SELECT user_guid FROM wallet WHERE wallet.guid = \"limit\".wallet_guid)")
             exec("ALTER TABLE \"limit\" DROP COLUMN wallet_guid")
             exec("ALTER TABLE \"limit\" ALTER COLUMN user_guid SET NOT NULL")
             exec("ALTER TABLE \"limit\" ADD CONSTRAINT unique_limit UNIQUE(market_guid, user_guid)")

```

imykolenko: Thanks a lot for taking care. I thought I verified specifically that migration is backward compatible, maybe there was an additional change in between..

backend/src/main/kotlin/xyz/funkybit/core/model/db/Wallet.kt
```diff
@@ -23,26 +25,58 @@ value class WalletId(override val value: String) : EntityId {
     override fun toString(): String = value
 }
 
+enum class WalletFamily {
```
antonyshchenko, line 28: This duplicates `NetworkType` enum we already had.

backend/src/main/kotlin/xyz/funkybit/core/model/db/WalletAuthorization.kt
```diff
@@ -0,0 +1,50 @@
...
+    val walletGuid = reference("wallet_guid", WalletTable).index()
+    val createdAt = timestamp("created_at")
+    val createdBy = varchar("created_by", 10485760)
+    val message = varchar("message", 10485760)
+    val signature = varchar("signature", 10485760)
+}
+
+class WalletAuthorizationEntity(guid: EntityID<WalletAuthorizationId>) : GUIDEntity<WalletAuthorizationId>(guid) {
+    companion object : EntityClass<WalletAuthorizationId, WalletAuthorizationEntity>(WalletAuthorizationTable) {
+        fun store(wallet: WalletEntity, createdBy: Address, message: String, signature: String): WalletAuthorizationEntity {
```
antonyshchenko, line 31: Nit: we usually call these kind of methods `create`. Would also be nice to pull some of the logic from `WalletRoutes` into the model. E.g it could be `Wallet.addForUser(...)` method that would create a new wallet, serialize the proof message and store the proof.

backend/src/main/kotlin/xyz/funkybit/core/model/db/Wallet.kt
imykolenko: yep, good catch, currently they are 1:1.

PR #468, opened 2024-08-30
UI: support for linking EVM wallet when connected with Bitcoin wallet

antonyshchenko:
- Currently wallet linking is triggered from the balances table, we might want to add it to other places in UI
- Wallet button in the header changes to the linked wallet address which is probably not what we want, I've not yet figured out how to change this, probably it has something to do with `primaryCategory` memo in `walletProvider`

Demo:

https://github.com/user-attachments/assets/5ef42eb8-fffc-4bed-9720-351cf715be1b


backend/src/main/kotlin/xyz/funkybit/core/evm/EIP712Helper.kt
```diff
@@ -117,7 +117,7 @@ object EIP712Helper {
                 mapOf(
                     "message" to linkMessage.message,
                     "address" to linkMessage.address,
-                    "linkedAddress" to linkMessage.linkAddress,
+                    "linkAddress" to linkMessage.linkAddress,
```
antonyshchenko, line 120: Took me a while to figure out why signature verification was failing

backend/src/main/kotlin/xyz/funkybit/core/model/Address.kt
```diff
@@ -156,7 +156,7 @@ sealed class BitcoinAddress(val value: String) : Address() {
         org.bitcoinj.core.Address.fromString(params, value)
 }
 
-@Serializable
+@Serializable(with = EvmAddressSerializer::class)
```
antonyshchenko, line 159: Lack of custom serializer was leading to bad requests when calling the endpoint from UI. By default it was serialized into an object, while UI was sending a string.

web-ui/src/contexts/walletProvider.tsx
```diff
@@ -102,22 +85,32 @@ function setGlobalPrimaryCategory(category: WalletCategory) {
   setGlobal('primaryCategory', category)
 }
 
+export function getGlobalPrimaryChainId(): number | null {
+  return getGlobal('primaryChainId')
+}
+function setGlobalPrimaryChainId(chainId: number | null) {
```
antonyshchenko, line 91: I've replaced `getGlobalPrimaryAccount` and `setGlobalPrimaryAccount` with primary chain id because evm account serialization was causing `TypeError: Converting circular structure to JSON`

web-ui/src/tradingSymbols.ts
```diff
@@ -1,5 +1,6 @@
 import TradingSymbol from 'tradingSymbol'
 import { ConfigurationApiResponse } from 'apiClient'
+import { bitcoinEnabled } from 'components/Screens/HomeScreen/swap/ConnectWallet'
```
bholzman, line 3: maybe `bitcoinEnabled` should move somewhere more "util"y ?

bholzman (approved):
This looks good to me, but note my comment here: https://github.com/funkybit-labs/chainring-monorepo/pull/467#discussion_r1740272969 - I think we might want to adjust the linking protocol. 

The "Wallet" button in the header changing is because of this code in walletProvider.tsx:
```
  const primaryCategory = useMemo(() => {
    return evmAccount?.status === 'connected'
      ? 'evm'
      : bitcoinAccount?.address
        ? 'bitcoin'
        : 'none'
  }, [evmAccount, bitcoinAccount])
  ```
  if the primary category is already 'bitcoin' it shouldn't change it to 'evm'.
  
  Besides on the Deposit button, we should also require them to link a different wallet family when the pair they are trying to swap has one side in that wallet family. I think probably the "Swap" button should change to either "Connect EVM Wallet" or "Connect Bitcoin Wallet" as appropriate.

backend/src/main/kotlin/xyz/funkybit/core/evm/EIP712Helper.kt
imykolenko: sorry for that. I also updated in both places to `authorizedAddress`.

imykolenko:
- wallets can authorize each other 
     - primary wallet signs authorization, authorized wallet signs regular sign-in and registers authorization
     - works both ways
 - wallet of primary category is displayed in header. On primary disconnect authorized wallet is used.
 - swap widget checks that all required categories are connected

web-ui/src/tradingSymbols.ts
imykolenko: moved to bitcoin context, seems a good fit.

PR #469, opened 2024-09-03
CHAIN-486 check bitcoin address matches signature

imykolenko:
Check that claimed address (sign-in, link messages) matches the address recovered from the signature. 

P2PKH verification is straightforward. 

P2SH is weird because we construct 1 of 1 scriptSig with the provided address and then check is script hash matches recovered value. Other multisig options will fail because we do not have original script.

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinSignatureVerification.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinSignatureVerification.kt
@@ -67,15 +67,19 @@ object BitcoinSignatureVerification {
                 }
 
                 is BitcoinAddress.P2SH -> {
-                    // create the 1 of 1 scriptSig: OP_HASH160 <20-byte hash160(pubkey)> OP_EQUAL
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/BitcoinSignatureVerification.kt
```diff
@@ -48,10 +54,34 @@ object BitcoinSignatureVerification {
                 BigInteger(1, signatureHex.substring(66, 130).hexToByteArray()),
             )
 
-            val pubkey =
-                ECKey.recoverFromSignature(recId, ECKey.ECDSASignature(r, s), Sha256Hash.of(msgHash), compressed)
```
imykolenko, line 52: `Sha256Hash.of` hashed 3rd time while wrapping, took a while to figure out

bholzman: whoops, sorry

bholzman (approved):
Looks good!

PR #474, opened 2024-09-06
CHAIN-475: testnet enrollment

bholzman:


bflood366 (approved):
looks great 👍

web-ui/src/components/Screens/HomeScreen/swap/ConnectWallet.tsx
```diff
@@ -41,7 +41,6 @@ export function ConnectWallet({ onSwitchToChain }: Props) {
               : `Connect ${missingConnection}Wallet`
           }}
           onClick={() => {
-            console.log('ConnectWallet', 'onClick')
```
imykolenko, line 44: 👌

imykolenko (approved):

backend/src/main/kotlin/xyz/funkybit/apps/api/model/Configuration.kt
```diff
@@ -28,6 +31,11 @@ data class AccountConfigurationApiResponse(
     val newSymbols: List<SymbolInfo>,
     val role: Role,
     val authorizedAddresses: List<Address>,
+    val testnetChallengeStatus: TestnetChallengeStatus,
+    val testnetChallengeDepositSymbol: String?,
+    val testnetChallengeDepositContract: Address?,
```
antonyshchenko, line 36: Nit: seems like this three fields should be grouped together into a data class.

web-ui/src/components/Screens/HomeScreen/TestnetChallengeTab.tsx
```diff
@@ -0,0 +1,189 @@
...
+        setTestnetChallengeDepositContract(symbolContract)
+        if (symbol.chainId != evmConfig.state.chainId) {
+          setSwitchToEthChainId(symbol.chainId)
+        }
+        setShowTestnetChallengeDepositModal(true)
+      }
+    }
+  }, [accountConfigQuery, evmConfig.state.chainId, symbols])
+
+  const autoDepositTriggeredRef = useRef(false)
```
antonyshchenko, line 97: Wondering why do we have to use refs here?

backend/src/main/kotlin/xyz/funkybit/apps/api/model/Configuration.kt
bholzman: you're right, that would be cleaner

web-ui/src/components/Screens/HomeScreen/TestnetChallengeTab.tsx
bholzman: we don't, ended up this way after some debugging, I'm changing it to use a more standard `useState` in my current branch

PR #478, opened 2024-09-09
bitcoin/arch deposits

bflood366:


contracts
bflood366, line 1: looks like some other PR reverted the contract submodule to an old one, so this just brings in the correct one again.

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/testutils/AppUnderTestRunner.kt
```diff
@@ -122,11 +127,25 @@ class AppUnderTestRunner : BeforeAllCallback, BeforeEachCallback {
...
+                                } catch (e: Exception) {
+                                    null
+                                }
+                                ArchAccountBalanceIndexTable.deleteAll()
+                                if (
+                                    programAccount?.status != ArchAccountStatus.Complete ||
+                                    ArchAccountEntity.findProgramStateAccount()?.status != ArchAccountStatus.Complete ||
+                                    onChainAccount == null ||
+                                    !onChainAccount.data.toByteArray().contentEquals(javaClass.getResource("/exchangeprogram.so")!!.readBytes())
+                                ) {
```
bflood366, line 142: this change is just useful for local testing - essentially on starting the tests if the program is already deployed on chain (and binary matches file in resources dir) and the state accounts are setup in DB, it won't delete those DB entries, so  when backend starts for this test run it won't need to deploy the program etc which saves a ton of time.

backend/src/main/kotlin/xyz/funkybit/core/model/db/ArchAccountBalanceIndex.kt
```diff
@@ -0,0 +1,156 @@
...
+}
+
+enum class ArchAccountBalanceIndexStatus {
+    Pending,
+    Assigning,
+    Assigned,
+    Failed,
+}
+
+object ArchAccountBalanceIndexTable : GUIDTable<ArchAccountBalanceIndexId>("arch_account_balance_index", ::ArchAccountBalanceIndexId) {
```
antonyshchenko, line 37: Does this table really need a guid primary key? Could we use a compound primary key (wallet_guid + arch_account_guid) instead? You can check `OrderBookSnapshotTable` for an example how to define a table without guid. `PrimaryKey` constructor allows providing multiple columns.

antonyshchenko (approved):
Nice job! 👍

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchChainWorker.kt
```diff
@@ -2,20 +2,33 @@ package xyz.funkybit.apps.ring
...
+
     fun start() {
         if (BitcoinClient.bitcoinConfig.enabled) {
             try {
-                ArchContractsPublisher.updateContracts()
+                thread(start = true, name = "arch_contract_deployer", isDaemon = false) {
+                    ArchContractsPublisher.updateContracts()
+                    depositHandler.start()
+                    blockchainTransactionHandler.start()
+                }
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -0,0 +1,207 @@
...
+                                if (index == -1) {
+                                    // throw an error since we did not find it
+                                    throw Exception("Did not find an index for wallet ${update.walletAddress.value}")
+                                }
+                                UpdateArchAccountBalanceIndexAssignment(
+                                    update.entity,
+                                    balances.size - index - 1,
+                                )
+                            }
+                        }.flatten()
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/xyz/funkybit/core/model/db/ArchAccountBalanceIndex.kt
bflood366: I am gonna leave as is for now. It looks like support for entities with a compound primary key was part of exposed 0.50.1 (we are 0.48.0). I tried to upgrade to that latest exposed (0.54.0) and got tons of compile errors, so looks like there was some interface changes - I will add a different ticket to upgrade our exposed and kotlin version since I didn't want to start down that path just for this.

PR #483, opened 2024-09-11
CHAIN-495 Add wallet ownership/authorization signatures as sequencer input

imykolenko:
EIP712 message is stored as json string including type definitions, this should allow to deal with message structure updates.


bholzman (approved):
Looks great!

PR #484, opened 2024-09-11
allow ws connection without wallet

imykolenko:


web-ui/src/contexts/websocket.tsx
```diff
@@ -122,7 +122,13 @@ export function WebsocketProvider({
 
     async function connect(refreshAuth: boolean = false) {
       // don't do anything until we know for sure that wallet is connected
-      if (wallet.primaryCategory === 'none') return
+      if (
+        wallet.primaryCategory === 'evm' &&
```
bholzman, line 126: nit: might be nice to encapsulate this logic into the wallet - `wallet.isConnecting()` or somesuch

bholzman (approved):
Looks good!

web-ui/src/contexts/websocket.tsx
antonyshchenko: @bholzman I've already addressed this in my branch

imykolenko: @antonyshchenko did you address websocket connect as well? Maybe just add this like and I will close PR

PR #485, opened 2024-09-11
Make primary wallet implicit in UI

antonyshchenko:
https://github.com/user-attachments/assets/adca85e9-f8f0-450b-96cc-e1ac422f5d9c

### DIFF
--- a/web-ui/src/components/Screens/Header.tsx
+++ b/web-ui/src/components/Screens/Header.tsx
@@ -1,6 +1,6 @@
 import logo from 'assets/funkybit-orange-logo-name.png'
 import { useWallets } from 'contexts/walletProvider'
-import { classNames, shortenedWalletAddress, uniqueFilter } from 'utils'
+import { classNames, abbreviatedWalletAddress, uniqueFilter } from 'utils'
 import { Button } from 'components/common/Button'
 import { Popover, Transition } from '@headlessui/react'
 import React, { Fragment, useEffect, useMemo, useState } from 'react'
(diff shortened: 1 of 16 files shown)


web-ui/src/components/Screens/Header.tsx
```diff
@@ -94,52 +69,104 @@ export function Header({
...
-                  )}
-                  {wallet.primaryCategory === 'evm'
-                    ? evmAddressDisplay(wallet.primaryAddress ?? '0x')
-                    : bitcoinAddressDisplay(wallet.primaryAddress)}
-                </span>
+            <Popover className="relative">
+              {() => (
+                <>
+                  <Popover.Button className="flex items-center gap-3 overflow-hidden text-ellipsis rounded-[20px] bg-darkBluishGray7 px-4 py-2 text-darkBluishGray1 transition-colors duration-300 ease-in-out hover:bg-primary4 focus:outline-none focus:ring-1 focus:ring-inset focus:ring-mutedGray">
+                    Connected wallets:
```
bholzman, line 78: I kind of think we can probably get away without saying anything here -- they just connected their wallet and there's a new header icon, I think people will figure out what it is for

bholzman (approved):
Looks great, nice cleanup!

web-ui/src/components/Screens/Header.tsx
antonyshchenko: Idk, a button with icons and no text could looks weird, but I will try tomorrow

antonyshchenko: Tried without text and I'm not convinced. Going to leave it as is. Here are some screenshots:

Screenshot 2024-09-12 at 12 35 11
Screenshot 2024-09-12 at 12 35 41
Screenshot 2024-09-12 at 12 35 54
Screenshot 2024-09-12 at 12 36 20

PR #486, opened 2024-09-11
CHAIN-477: leaderboard ui

bholzman:
Screenshot 2024-09-11 at 2 47 56 PM


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -149,7 +155,12 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
                 lastPage = maxPage.toInt(),
                 entries.map { entry ->
                     LeaderboardEntry(
-                        entry[UserTable.nickName] ?: entry[WalletTable.address],
+                        entry[UserTable.nickName] ?: Address.auto(entry[WalletTable.address]).let {
+                            when (it) {
+                                is EvmAddress -> it.abbreviated()
+                                is BitcoinAddress -> it.abbreviated()
+                            }
```
imykolenko, line 162: It would make sense to introduce `abbreviated()` on Address

imykolenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
bholzman: indeed, added

PR #488, opened 2024-09-12
CHAIN-479 testnet points distribution

imykolenko:


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -112,6 +126,88 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
             )
         }
 
+        fun distributePoints(challengePNLType: TestnetChallengePNLType) {
```
imykolenko, line 129: potentially this also can be done as one native query (insert from select) to avoid insert per user. In this case should generate id in the database.

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -112,6 +126,88 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
...
+            val rankDef = rowNumber().over().partitionBy(TestnetChallengePNLTable.type).orderBy(TestnetChallengePNLTable.currentBalance, SortOrder.DESC).alias("rank")
```
bholzman, line 130: we should be ordering by (currentBalance - initialBalance) / initialBalance, just like we do in `getLeaderboard` - it's what percentage profit you've made over the period that counts, not your absolute balance.

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
bholzman: I think that would probably be best

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -112,6 +126,88 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
...
+                .select(TestnetChallengePNLTable.userGuid, TestnetChallengePNLTable.type, rankDef, totalUsersDef)
+                .where { TestnetChallengePNLTable.type.eq(challengePNLType) }
+
+            rankedUsers.forEach { row ->
+                val userGuid = row[TestnetChallengePNLTable.userGuid]
+                val rank = row[rankDef]
+                val totalUsers = row[totalUsersDef]
+                val timeGroup = row[TestnetChallengePNLTable.type]
+
+                val points = when {
```
bholzman, line 143: FYI in https://github.com/funkybit-labs/chainring-monorepo/pull/491 I added a "reward_category" column to the user reward table so the UI knows if the user was in the top 25% or bottom 5% or whatever.

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -120,7 +216,7 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
 
             val rowsPerPage = 20
             val maxPage = ceil(count.div(rowsPerPage.toDouble())).toLong()
-            // ensure page is sane
+            // ensure page is same
```
bholzman, line 219: no, sane is what I meant -- not outside the bounds, and zero-indexed

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
imykolenko: makes sense let user know 👍 

My assumption is that on the week change user is granted both daily reward and weekly rewards. Should `reward_category` be always set to daily rewards? Or maybe `reward_category` should represent highest order event occurred, e.g. (weekly, if not - daily)

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
imykolenko: Apologies for the confusion!

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
imykolenko: fixed

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
imykolenko: Updated to calculate ranking and grant reward points in one database query.

Btw, I found implementation of typeId for postgres, it is added as migration (uuid_v7, base32, typeId). Not strictly necessary, but nice that id format stays.

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
imykolenko: I moved rewards granting to native sql query. Storing also `reward_category` is very simple change, will do it right after you merge https://github.com/funkybit-labs/chainring-monorepo/pull/491.

backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V87_TypeId.kt
```diff
@@ -0,0 +1,319 @@
+package xyz.funkybit.core.model.db.migrations
+
+import org.jetbrains.exposed.sql.transactions.transaction
+import xyz.funkybit.core.db.Migration
+
+@Suppress("ClassName")
+class V87_TypeId : Migration() {
+
+    override fun run() {
+        // https://github.com/jetify-com/typeid-sql
```
bholzman, line 10: very cool, I'm sure will come in handy later too

bholzman (approved):
Looks great!

PR #489, opened 2024-09-12
Handle wallet authorization failures in UI

antonyshchenko:
https://github.com/user-attachments/assets/9830c912-8511-4c4d-9058-ec84be35c80e

### DIFF
--- a/web-ui/src/auth.ts
+++ b/web-ui/src/auth.ts
@@ -5,6 +5,8 @@ import {
   getGlobalPrimaryChainId,
   getGlobalPrimaryAddress
 } from 'contexts/walletProvider'
+import { bitcoinSignMessage } from 'contexts/bitcoin'
+import { UserRejectedRequestError } from 'viem'
 
 export type LoadAuthTokenOptions = {
   forceRefresh: boolean
@@ -23,6 +25,13 @@ export async function loadAuthToken(
 
     if (!signingPromise) {
       signingPromise = signAuthToken(primaryAddress, primaryChainId)
+        .then((maybeToken) => {
+          if (maybeToken == null) {
+            throw Error('Rejected')
+          } else {
+            return maybeToken
+          }
+        })
         .then((authToken) => {
           signingPromise = null
           return authToken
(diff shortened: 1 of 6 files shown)


web-ui/src/auth.ts
```diff
@@ -23,6 +25,13 @@ export async function loadAuthToken(
 
     if (!signingPromise) {
       signingPromise = signAuthToken(primaryAddress, primaryChainId)
+        .then((maybeToken) => {
+          if (maybeToken == null) {
+            throw Error('Rejected')
```
antonyshchenko, line 30: We get here if user rejects auth token signing. Currently we would propose to sign the token a few times and then give up leaving the app in broken state. We don't have any explicit logic for this and I think that it might be a side-effect of React Query retrying failed requests by default.

I think the better way to handle this would be to log them out by disconnecting wallets. Thoughts? In any case I will address this with a separate PR.

web-ui/src/auth.ts
```diff
@@ -42,7 +51,7 @@ export async function loadAuthToken(
 export async function signAuthToken(
   address: string,
   chainId: number
-): Promise<string> {
+): Promise<string | null> {
```
antonyshchenko, line 54: `null` here represents rejection by user

web-ui/src/auth.ts
bholzman: That makes sense to me — if they reject the auth token signing we should just disconnect their wallet

web-ui/src/contexts/walletProvider.tsx
```diff
@@ -289,58 +280,81 @@ function WalletProviderInternal({ children }: { children: React.ReactNode }) {
...
+              chainId: chainId,
+              signature: evmSignature,
+              timestamp: timestamp
+            }
+          )
+        } catch (error) {
+          console.log('Error during wallet authorization', error)
+          bitcoinWallet.disconnect()
+          alert(
+            'Something went wrong while authorizing a wallet, please try again or contact customer support'
```
bholzman, line 352: instead of generic "customer support" should give them I guess our discord - https://discord.gg/pzYr3mdU3D -- although obviously they can't click on it from the alert, so maybe better to just say "please try again or reach out on discord"

bholzman (approved):
Looks great!

web-ui/src/contexts/walletProvider.tsx
antonyshchenko: Good point, changed the message

PR #491, opened 2024-09-13
CHAIN-508: CTAs and other messages as "cards" in challenge UI

bholzman:
https://github.com/user-attachments/assets/d9a808e6-45c6-4c18-b2d7-9672c258c977


imykolenko (approved):
looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/Withdrawal.kt
```diff
@@ -244,6 +244,12 @@ class WithdrawalEntity(guid: EntityID<WithdrawalId>) : GUIDEntity<WithdrawalId>(
                 execute(TransactionManager.current())
             }
         }
+
+        fun existsForUserAndNetworkType(user: UserEntity, networkType: NetworkType): Boolean {
+            return WithdrawalTable.innerJoin(WalletTable).select(WithdrawalTable.id).where {
+                WalletTable.networkType.eq(networkType) and WalletTable.userGuid.eq(user.guid)
+            }.limit(1).count() == 1L
```
imykolenko, line 251: nit: query already limits to 1, checking if empty can be simplified to `.limit(1).any()`

backend/src/main/kotlin/xyz/funkybit/apps/api/model/TestnetChallenge.kt
```diff
@@ -24,3 +24,18 @@ data class LeaderboardEntry(
...
+    BitcoinConnect,
+    BitcoinWithdrawal,
+    EvmWithdrawal,
+}
+
+@Serializable
+data class Card(
+    val type: CardType,
+    val params: Map<String, String>,
+)
```
antonyshchenko, line 41: Just curious: why like this and not a sealed class hierarchy?

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/testnetchallenge/TestnetChallengeTest.kt
```diff
@@ -1,54 +1,73 @@
...
 
+data class Trader(
+    val apiClient: TestApiClient,
+    val wallet: Wallet,
+    val wsClient: WsClient,
+) {
+    val a = apiClient
+    val w = wallet
+    val ws = wsClient
+}
```
antonyshchenko, line 45: 👍 Would be nice to move it to `testutils` package though.

backend/src/main/kotlin/xyz/funkybit/core/model/db/Withdrawal.kt
bholzman: sure, can change

backend/src/main/kotlin/xyz/funkybit/apps/api/model/TestnetChallenge.kt
bholzman: ah, no reason, a sealed class hierarchy is better, I'll change it

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/testnetchallenge/TestnetChallengeTest.kt
bholzman: I added this here to simplify the return value of the `enrollInTestnetChallenge()` helper, but I figured that CHAIN-484 will change it anyway, so didn't want to necessarily propagate this.

PR #492, opened 2024-09-13
Eco mode for Sequencer to preserve battery while running locally

antonyshchenko:


bholzman (approved):
Nice!

imykolenko (commented):
👍 
thanks for taking care!

PR #493, opened 2024-09-13
Waf initial attempt

sburketx:
Initial attempt of the waf.


terraform/modules/waf/main.tf
```diff
@@ -0,0 +1,399 @@
...
+    name     = "${var.short_env_name}AWSRateBasedRuleDOS"
+    priority = 5
+
+    action {
+      block {}
+    }
+
+    statement {
+      rate_based_statement {
+        limit              = 500
```
bholzman, line 116: we have a separate dev ticket to put rate limits in, I think 500 over 5 minutes is definitely too low. we can leave the rule here as a placeholder, but let's change to, say, 5000.

terraform/testnet/main.tf
```diff
@@ -102,6 +102,18 @@ module "web" {
   certificate_arn = data.aws_acm_certificate.funkybit_fun_us_east_1.arn
 }
 
+module "waf" {
+  source                   = "../modules/waf"
+  aws_region               = var.aws_region
+  aws_lb_arn               = module.alb.https_listener_arn
+  short_env_name           = local.name_prefix
+  ipv4_blacklist_addresses = local.ipv4blacklist
+  ipv4_whitelist_addresses = local.ipv4whitelist
```
bholzman, line 111: I think we'll need to whitelist the mocker and telegram bot container ips here, won't do to have our market maker getting blocked!

bholzman (approved):
Looks good!

terraform/modules/waf/main.tf
sburketx: Done. I've moved the setting to the variables.tf under testnet for ease of adjustment.

terraform/testnet/main.tf
sburketx: the telegram bot container I've just accounted for and will push the changes shortly. In the above code, I used the test-api lb which appears to be identical to the mocker-lb , lemme know if I have misunderstood. Does the bastion or the webapp need whitelisting?

PR #495, opened 2024-09-16
CHAIN-481 referral points

imykolenko:
- added "copy invite code" icon with tooltip to follow design
   - atm code redeem is silent, let's come up with the way how to inform user
- referral bonus is 10%
- referral points are granted inside one DB call all the invite chain up until bonus size is too small (0.01 point)
 

====
Screenshot 2024-09-16 at 3 15 27 PM

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
@@ -78,7 +78,7 @@ class TestnetChallengeUserRewardEntity(guid: EntityID<TestnetChallengeUserReward
                             r.${TestnetChallengeUserRewardTable.userGuid.name} AS invitee_guid,
                             r.${TestnetChallengeUserRewardTable.amount.name}::numeric AS invitee_points,
                             u.${UserTable.invitedBy.name} AS invitor_guid,
-                            r.${TestnetChallengeUserRewardTable.amount.name} * $referralBonusSize AS invitor_points,        -- 'referralBonusFraction' of the previous bonus
+                            r.${TestnetChallengeUserRewardTable.amount.name} * $referralBonusSize AS invitor_points,        -- 'referralBonusSize' of the previous bonus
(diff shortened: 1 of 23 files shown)


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
```diff
@@ -65,21 +64,64 @@ object TestnetChallengeUserRewardTable : GUIDTable<TestnetChallengeUserRewardId>
...
-            }
+        fun distributeReferralPoints(
+            earnedAfter: Instant,
+            earnedBefore: Instant,
+            referralBonusSize: BigDecimal,
+            referralBonusMinAmount: BigDecimal,
+        ) {
+            TransactionManager.current().exec(
+                """
+                    WITH RECURSIVE referral_chain(invitee_guid, invitee_points, invitor_guid, invitor_points, level) AS (
```
bholzman, line 75: This is very cool, but I don’t think we need to do this. If we just grant reward points to the immediate referrer, whoever referred that person would still effectively get their 1% bonus, just delayed by 1 day.

imykolenko: I would rather to leave it in place, there is only one assignment per user collected from the invite chain so amount of work is not really different. Also it is slightly better UX to receive earned points immediately.

imykolenko: So bonus should be 1%. What should be minimal value which isn't worth granting?

bholzman: that's not what I meant -- the bonus is 10% ... so if person A refers person B who then refers person C, for every 100 points C earns, B should earn 10. The 1% thing is just saying that person A effectively receives 1% of what C earns by virtue of earning 10% of what B earns.

bholzman (approved):
Looks great!

PR #496, opened 2024-09-16
settlements on arch

bflood366:


integrationtests/build.gradle.kts
```diff
@@ -86,6 +86,7 @@ tasks.test {
         "--add-opens=java.base/java.io=ALL-UNNAMED",
         "--add-opens=java.base/java.util=ALL-UNNAMED"
     )
+    maxHeapSize = "768m"
```
bflood366, line 89: this increases the heap size when running unit tests - I was seeing some OOM exceptions and the default is 512m

backend/src/main/kotlin/xyz/funkybit/apps/api/services/ExchangeApiService.kt
```diff
@@ -276,6 +300,25 @@ class ExchangeApiService(
...
+        }
+        val bitcoinOrderMessage = "[funkybit] Please sign this message to authorize a swap. This action will not cost any gas fees." +
+            if (orderRequest.side == OrderSide.Buy) {
+                "\nSwap $amount ${quoteSymbol.name} for ${baseSymbol.name}"
+            } else {
+                "\nSwap $amount ${baseSymbol.name} for ${quoteSymbol.name}"
+            } + when (orderRequest) {
+                is CreateOrderApiRequest.Limit -> "\nPrice: ${orderRequest.price.toFundamentalUnits(quoteSymbol.decimals)}"
+                else -> "\nPrice: Market"
+            } + "\nAddress: ${bitcoinAddress.value}, Nonce: ${orderRequest.nonce}"
```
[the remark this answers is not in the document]
bholzman: I think we can probably just keep the chains in the symbol. It's also probably fine to keep the nonce.

backend/src/main/kotlin/xyz/funkybit/apps/ring/SettlementCoordinator.kt
```diff
@@ -59,7 +73,7 @@ class SettlementCoordinator(
 ) {
     private val marketMap = mutableMapOf<MarketId, MarketEntity>()
     private val symbolMap = mutableMapOf<SymbolId, SymbolEntity>()
-    private val chainIds = blockchainClients.map { it.chainId }
+    private val chainIds = blockchainClients.map { it.chainId } + listOf(BitcoinClient.chainId)
```
bholzman, line 76: should this look at the BITCOIN_NETWORK_ENABLED env var?

bholzman (approved):
Looks great, nice work!

backend/src/main/kotlin/xyz/funkybit/apps/api/services/ExchangeApiService.kt
```diff
@@ -286,8 +329,14 @@ class ExchangeApiService(
...
         val userId = wallet.userGuid.value
-        val walletAddress = wallet.address
         val symbol = getSymbolEntity(apiRequest.symbol)
+        val walletAddress = transaction {
+            if (wallet.networkType == symbol.chain.networkType) {
+                wallet.address
+            } else {
+                wallet.authorizedWallet(symbol.chain.networkType)?.address
+            } ?: throw RequestProcessingError("No authorized wallet found for network type for this symbol")
+        }
```
antonyshchenko, line 339: This could be extracted into a private method and reused in few places

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -60,23 +66,62 @@ class ArchTransactionHandler(
...
             logger.debug { "Arch Transaction handler thread starting" }
-            val (programPubkey, programBitcoinAddress) = transaction {
-                Pair(
-                    ArchAccountEntity.findProgramAccount()!!.rpcPubkey(),
-                    DeployedSmartContractEntity.programBitcoinAddress(),
-                )
+            transaction {
+                programAccount = ArchAccountEntity.findProgramAccount()!!
+                programStateAccount = ArchAccountEntity.findProgramStateAccount()!!
+                programBitcoinAddress = DeployedSmartContractEntity.programBitcoinAddress()
```
antonyshchenko, line 80: Won't this break if program is not deployed yet? Or do we call `start` after we make sure that program is deployed?

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -60,23 +66,62 @@ class ArchTransactionHandler(
...
+            var settlementBatchInProgress = false
+            var withdrawalBatchInProgress = false
+
             while (true) {
                 try {
                     transaction {
-                        processBalanceIndexBatch(programPubkey)
-                        processDepositBatch(programPubkey)
-                        processWithdrawalBatch(programBitcoinAddress, programPubkey)
+                        if (tryAcquireAdvisoryLock(chainId.value.toLong())) {
```
antonyshchenko, line 95: I might have missed it when it was added `BlockchainTransactionHandler`, but why do we need a lock here and in `BlockchainTransactionHandler`?

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -121,12 +166,14 @@ class ArchTransactionHandler(
...
 
-    private fun processBalanceIndexBatch(programPubkey: ArchNetworkRpc.Pubkey): Boolean {
+    private fun processBalanceIndexBatch(): Boolean {
+        val programPubkey = programAccount.rpcPubkey()
         val assigningBalanceIndexUpdates = ArchAccountBalanceIndexEntity.findAllForStatus(ArchAccountBalanceIndexStatus.Assigning)
         return if (assigningBalanceIndexUpdates.isNotEmpty()) {
             assigningBalanceIndexUpdates.firstOrNull()?.entity?.archTransaction?.let { archTransaction ->
                 ArchNetworkClient.getProcessedTransaction(archTransaction.txHash!!)?.let { processedTx ->
                     if (processedTx.status == ArchNetworkRpc.Status.Processed) {
+                        Thread.sleep(2000)
```
antonyshchenko, line 176: Is this a workaround for Arch bugs?

antonyshchenko (commented):
Nice job! 👍

backend/src/main/kotlin/xyz/funkybit/apps/ring/SettlementCoordinator.kt
bflood366: added.

backend/src/main/kotlin/xyz/funkybit/apps/api/services/ExchangeApiService.kt
bflood366: done, thx

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
bflood366: this is only started if the program has been deployed and program state account initialized

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
bflood366: yes

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
bflood366: we probably actually could remove from both places - it was initially added to prevent a case where if 2 instances of say the ring app started and there were 2 instances of the handler running if would prevent them both from processing at the same time since they would need to get the advisory lock for a given chain. I think subsequently there were some changes to the ecs_deploy to make sure the ring app is stopped first before another one is started.

backend/src/main/kotlin/xyz/funkybit/apps/api/services/ExchangeApiService.kt
```diff
@@ -166,6 +175,14 @@ class ExchangeApiService(
         val market = getMarket(batchOrdersRequest.marketId)
         val baseSymbol = getSymbolEntity(market.baseSymbol)
         val quoteSymbol = getSymbolEntity(market.quoteSymbol)
+        listOf(baseSymbol, quoteSymbol).forEach {
+            if (wallet.networkType != it.chainId.value.networkType()) {
+                transaction {
```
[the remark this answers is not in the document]
bflood366: I changed it a bit here to look at the set of network types, so if the wallet network type is bitcoin, but say both symbols are evm, it will only check once. We still will only open a transaction if they are different, if they are same then no transaction is opened.

PR #497, opened 2024-09-17
Limit deposits for testnet challenge participants

antonyshchenko:
Created a separate ticket to take withdrawals into account when distributing points: https://funkybit.atlassian.net/browse/CHAIN-518

### DIFF
--- a/web-ui/src/components/Screens/HomeScreen/DepositModal.tsx
+++ b/web-ui/src/components/Screens/HomeScreen/DepositModal.tsx
@@ -20,11 +20,13 @@ import { encodeFunctionData, EncodeFunctionDataParameters } from 'viem'
 import { isErrorFromAlias } from '@zodios/core'
 import TradingSymbol from 'tradingSymbol'
 import { ExpandableValue } from 'components/common/ExpandableValue'
+import { minBigInt } from 'utils'
 
 export default function DepositModal({
   exchangeContractAddress,
   walletAddress,
   symbol,
+  testnetChallengeDepositLimit,
   isOpen,
   close,
   onClosed,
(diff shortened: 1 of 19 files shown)


web-ui/src/components/Screens/HomeScreen/DepositModal.tsx
```diff
@@ -162,15 +165,18 @@ export default function DepositModal({
 
     try {
       if (walletBalanceQuery.status == 'success') {
-        const availableAmount = walletBalanceQuery.data
+        const availableAmount =
+          testnetChallengeDepositLimit !== undefined
+            ? minBigInt(testnetChallengeDepositLimit, walletBalanceQuery.data)
```
antonyshchenko, line 170: Will prevent form submission without telling the reason to the user. Should we inform the user about the limit somehow?

bholzman: yes, we should definitely inform the user
how about this:
 - if their deposit limit is zero, the Deposit button is disabled with a tooltip that says "Deposits are limited during the Testnet Challenge to previously withdrawn assets."
 - and then in the deposit modal, when the deposit limit exists (even if 0) say something like: "During the Testnet Challenge, deposits are limited to previously withdrawn assets. Current deposit limit: 5,000"

bholzman (approved):
Looks great!

web-ui/src/components/Screens/HomeScreen/DepositModal.tsx
antonyshchenko: Done, here are screenshots:

Screenshot 2024-09-18 at 13 24 11
Screenshot 2024-09-18 at 13 24 19
Screenshot 2024-09-18 at 13 24 48

PR #498, opened 2024-09-17
CHAIN-509 grant points on specific actions

imykolenko:


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
```diff
@@ -124,25 +144,55 @@ class TestnetChallengeUserRewardEntity(guid: EntityID<TestnetChallengeUserReward
...
+            val rewardType = when (wallet.address) {
+                is BitcoinAddress -> TestnetChallengeUserRewardType.BitcoinWithdrawalDeposit
+                is EvmAddress -> TestnetChallengeUserRewardType.EvmWithdrawalDeposit
+            }
+
+            // both withdrawal and deposit are required for the award.
+            createIfNotExist(user, wallet, rewardType, prerequisites = { WithdrawalEntity.existsCompletedForWallet(wallet) })
+        }
+
+        private fun createIfNotExist(user: UserEntity, wallet: WalletEntity, rewardType: TestnetChallengeUserRewardType, prerequisites: (WalletEntity) -> Boolean = { true }) {
```
antonyshchenko, line 175: Nit: s is missing in the end `createIfNotExists`

backend/src/main/kotlin/xyz/funkybit/core/model/db/Deposit.kt
```diff
@@ -242,6 +247,8 @@ class DepositEntity(guid: EntityID<DepositId>) : GUIDEntity<DepositId>(guid) {
         val now = Clock.System.now()
         this.updatedAt = now
         this.status = DepositStatus.Complete
+
+        TestnetChallengeUserRewardEntity.createDepositReward(this.wallet.user, this.wallet)
```
antonyshchenko, line 251: Unless I'm missing something, this should be wrapped into a check for user testnet challenge status: 
```if (user.testnetChallengeStatus == TestnetChallengeStatus.Enrolled)```

File: backend/src/main/kotlin/xyz/funkybit/core/model/db/Wallet.kt [hunk 3]
```diff
@@ -73,6 +73,10 @@ class WalletEntity(guid: EntityID<WalletId>) : GUIDEntity<WalletId>(guid) {
                 }
                 this.user = user
             }
+
+            TestnetChallengeUserRewardEntity.createWalletConnectedReward(user, wallet)
```
antonyshchenko, line 77: Consider checking `user.testnetChallengeStatus`

backend/src/main/kotlin/xyz/funkybit/core/model/db/Withdrawal.kt
```diff
@@ -276,6 +282,10 @@ class WithdrawalEntity(guid: EntityID<WithdrawalId>) : GUIDEntity<WithdrawalId>(
         responseSequence?.let {
             this.responseSequence = responseSequence
         }
+
+        if (status == WithdrawalStatus.Complete) {
+            TestnetChallengeUserRewardEntity.createWithdrawalReward(this.wallet.user, this.wallet)
```
antonyshchenko, line 287: Consider checking `user.testnetChallengeStatus`

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/Withdrawal.kt
imykolenko: you right, at this point user should be already enrolled

imykolenko: hm. Being enrolled in challenge means participation in the leaderboard for earning performance based points. 
Here we are just granting points for the deposit/withdrawal using every wallet unconditionally. Once users will be enrolled - points it will be shown in UI. Otherwise it is more complex to track that deposit/withdrawal happened only after the enrollment. 

@bholzman should additional points be granted only to enrolled users?

bholzman: I think it’s probably ok if we are granting points on specific actions to everybody even if they haven’t enrolled in the testnet challenge, as long as we don’t include them in the leaderboard. We probably ought to fail enrollment (move them immediately to disqualified if they try to enroll) if the user has already made any other deposits though

imykolenko: Right, users who decide to enroll later still get all benefits for already performed actions. 

Good point to filter based on enrollment status when granting leaderboard rewards, and also fail enrollment due to prior deposits in challenge's symbol. Thx.

backend/src/main/kotlin/xyz/funkybit/core/model/db/Deposit.kt
imykolenko: as discussed below I will leave it as is. Users who join challenge later will still receive all benefits for already performed actions.

backend/src/main/kotlin/xyz/funkybit/core/model/db/Withdrawal.kt
imykolenko: addressed both

PR #499, opened 2024-09-18
CHAIN-507: backend support for expanded back-to-backs

bholzman:
Still could use some cleanup but the existing and new test cases pass
This is just the sequencer changes, there are UI changes needed too, will be in a separate PR


sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/SequencerApp.kt
```diff
@@ -520,26 +558,105 @@ class SequencerApp(
...
+                OrderSide.Sell -> Pair(
+                    calculateAmountForPercentageSell(
+                        firstMarket,
+                        account,
+                        order.percentage,
+                    ),
+                    null,
+                )
+                OrderSide.Buy -> {
+                    val (amount, maxAvailable) = calculateAmountForPercentageBuy(
```
bflood366, line 573: I don't think you need these local variables here

bflood366 (approved):
Looks great - awesome job 👍

antonyshchenko (commented):
Great idea to introduce `BaseAmount` and `QuoteAmount` types 👍

PR #501, opened 2024-09-18
remove mempool dependency and manage utxos internally

bflood366:


bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinBlockProcessor.kt
```diff
@@ -0,0 +1,242 @@
...
+                    logger.error(e) { "Unhandled exception in bitcoin block processor" }
+                }
+            }
+        }
+    }
+
+    fun processBlocks() {
+        if (transaction { ChainEntity.findById(chainId) } == null) {
+            return
+        }
```
antonyshchenko, line 65: Nit: we could stop checking this once chain record is created

backend/src/main/kotlin/xyz/funkybit/core/blockchain/ChainManager.kt
```diff
@@ -134,7 +134,7 @@ object ChainManager {
         blockExplorerUrl = System.getenv("BLOCK_EXPLORER_URL_BITCOIN") ?: "http://localhost:1080",
         faucetAddress = BitcoinAddress.canonicalize("bcrt1q3nyukkpkg6yj0y5tj6nj80dh67m30p963mzxy7"),
         submitterPrivateKey = (System.getenv("BITCOIN_SUBMITTER_PRIVATE_KEY") ?: "0x7ebc626d01c2d916c61dffee4ed2501f579009ad362360d82fcc30e3d8746cec").toHexBytes(),
-        feePayerPrivateKey = (System.getenv("BITCOIN_FEE_PAYER_PRIVATE_KEY") ?: ECKey().privateKeyAsHex).toHexBytes(),
+        feePayerPrivateKey = (System.getenv("BITCOIN_FEE_PAYER_PRIVATE_KEY") ?: "cec2c679ef0d11820c75b17174a8635d7aa1ce740b0bf32d80967c3bf7b3676d").toHexBytes(),
```
antonyshchenko, line 137: Do we want to commit this change?

backend/src/main/kotlin/xyz/funkybit/core/services/UtxoSelectionService.kt
```diff
@@ -1,139 +1,44 @@
...
-        return inputsSelector.selectInputs(
-            amount,
-            unspentUtxos.filter { it.reservedBy == null },
-            fee,
-        )
+    fun addUtxoAddressToMonitor(address: BitcoinAddress) {
+        transaction {
+            BitcoinUtxoAddressMonitorEntity.findById(BitcoinUtxoAddressMonitorId(address.value))
+                ?: BitcoinUtxoAddressMonitorEntity.create(address)
+        }
```
antonyshchenko, line 21: Since monitoring is not done by `UtxoSelectionService`, I would suggest moving this out. Could be just a method on `BitcoinUtxoAddressMonitorEntity`, e.g:

```
BitcoinUtxoAddressMonitorEntity.createIfNotExits(address)
```

Or alternatively this could be a static method of `BitcoinBlockProcessor`

backend/src/main/kotlin/xyz/funkybit/core/blockchain/ChainManager.kt
bflood366: I will, I changed the airdrop to always air drop the same amount - the fee payer is only used to pay the bitcoin fees during deployment of the contract so chances of stepping on each other is small and anyway we should be going to our local bitcoin node soon hopefully.

backend/src/main/kotlin/xyz/funkybit/core/services/UtxoSelectionService.kt
bflood366: I added the `createIfNotExists` to  `BitcoinUtxoAddressMonitorEntity` and removed this method from `UtxoSelectionService`

PR #503, opened 2024-09-20
CHAIN-492: dont send outputs for withdrawal batch, connecting token a…

bflood366:
…nd program state accounts etc


bholzman (approved):
Looks good!

PR #505, opened 2024-09-24
CHAIN-512: support for multiple token accounts

bflood366:


integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchSettlementTest.kt
```diff
@@ -407,4 +418,204 @@ class ArchSettlementTest : OrderBaseTest() {
...
+        Assumptions.assumeFalse(true)
+        val market = btcbtcArchMarket!!
+        val baseSymbol = btc
+        val quoteSymbol = btcArch
+
+        transaction {
+            ArchAccountTable.deleteWhere {
+                symbolGuid.isNotNull() and status.eq(ArchAccountStatus.Full)
+            }
+            KeyValueStore.setLong("TOKEN_ACCOUNT_SIZE_THRESHOLD", 50)
```
bholzman, line 433: ah, now I see why you put it in the DB but seems a bit like the tail wagging the dog. I thought for a bit but don't have an obviously better solution ... could just try to cover it in unit tests I suppose

Or perhaps we should add support in keyvaluestore to get cached values with a NOTIFY listener to reset the cache

bholzman (approved):
Looks good!

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchSettlementTest.kt
bflood366: I like the latter solution of getting cached values in KVStore with a listener - I will add a ticket for this but will leave as for now - agreed its not ideal whats there not, but felt it was important to be able to test this E2E with contract etc.

bflood366: actually I came up with a different way for now that does not require the DB which will work when integration tests run locally or on CI locally but not for deployed. I just made the value a var in ArchUtils which is an object so it shared by tests and backend when running locally.

bholzman: I had thought of that but realized that wouldn’t work in a ci integration test run, only locally

PR #513, opened 2024-09-25
Chain 532/update to arch v021

bflood366:


backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchContractsPublisher.kt
```diff
@@ -219,6 +276,18 @@ object ArchContractsPublisher {
...
     }
 
+    private fun createDeployedContractEntity(programAccount: ArchAccountEntity) {
+        val address = ArchNetworkClient.getAccountAddress(programAccount.rpcPubkey())
+        BitcoinUtxoAddressMonitorEntity.createIfNotExists(address)
+        DeployedSmartContractEntity.create(
+            name = ContractType.Exchange.name,
+            chainId = BitcoinClient.chainId,
+            implementationAddress = address,
+            proxyAddress = address,
```
bholzman, line 286: just idly wondering if we ought to allow proxyAddress to be null instead of just duplicating the implementation address

bholzman (approved):

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchContractsPublisher.kt
bflood366: yeah makes sense since there isn't a proxy in this case  - I will add a ticket for this

PR #517, opened 2024-09-26
CHAIN-521 disconnect wallet on address change

imykolenko:
- On EVM wallet address change:
  - If the EVM wallet is the primary, disconnect all wallets
  - If the Bitcoin wallet is the primary, disconnect only the EVM wallet

### DIFF
--- a/web-ui/src/contexts/walletProvider.tsx
+++ b/web-ui/src/contexts/walletProvider.tsx
@@ -244,6 +244,19 @@ function WalletProviderInternal({ children }: { children: React.ReactNode }) {
   // handle EVM wallet connect/disconnect events
   useEffect(() => {
     async function onConnected() {
+      // disconnect all wallets if the primary evm wallet address has changed
+      if (
+        primaryWallet &&
+        primaryWallet.networkType === 'Evm' &&
+        primaryWallet.address !== evmAccount.address
+      ) {
+        await disconnectWallet('Evm')
+
+        if (bitcoinAccount != null) {
+          await disconnectWallet('Bitcoin')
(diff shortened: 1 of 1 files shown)


web-ui/src/contexts/walletProvider.tsx
```diff
@@ -244,6 +244,19 @@ function WalletProviderInternal({ children }: { children: React.ReactNode }) {
   // handle EVM wallet connect/disconnect events
   useEffect(() => {
     async function onConnected() {
+      // disconnect all wallets if the primary evm wallet address has changed
+      if (
+        primaryWallet &&
+        primaryWallet.networkType === 'Evm' &&
+        primaryWallet.address !== evmAccount.address
+      ) {
```
antonyshchenko, line 252: Perhaps I'm missing something, but where is the `If Bitcoin wallet is the primary, disconnect only the EVM wallet` part from the PR description?

antonyshchenko: I mean, if Bitcoin wallet is the primary, we won't get into the logic inside this if-condition

imykolenko: I described both cases to explain all scenarios. In fact new code I introduced handles only case when EVM wallet is primary. 

Case when Bitcoin is primary is already covered: change of evm address triggers authorization, which results into message  `Can't connect another ${walletToAuthorizeNetworkType} wallet. You have been using wallet with address ${previouslyAuthorizedAddress} previously` and disconnect.

antonyshchenko (approved):

PR #519, opened 2024-09-26
introduce faucet address

imykolenko:
Introduce separate faucet key to be used instead of contract management key

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
@@ -510,3 +510,5 @@ class AirdropperBlockchainClient(config: BlockchainClientConfig) : BlockchainCli
         )
     }
 }
+
+class FaucetBlockchainClient(config: BlockchainClientConfig) : BlockchainClient(config)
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
```diff
@@ -510,3 +510,5 @@ class AirdropperBlockchainClient(config: BlockchainClientConfig) : BlockchainCli
         )
     }
 }
+
+class FaucetBlockchainClient(config: BlockchainClientConfig) : BlockchainClient(config)
```
antonyshchenko, line 514: I'm not sure what's the purpose of inheritance here.

If we want to distinguish the clients by role, then perhaps it would be better if `FaucetBlockchainClient` and `AirdropperBlockchainClient` would not inherit anything, but instead would contain an instance of `BlockchainClient` and expose only the relevant methods like `testnetChallengeAirdrop` and `faucetTransfer`.

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
imykolenko: makes sense, thx. Updated.

antonyshchenko (approved):

PR #523, opened 2024-09-27
Improve robustness of background worker threads

antonyshchenko:


backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -71,7 +71,12 @@ class ArchTransactionHandler(
...
-        archWorkerThread = thread(start = true, name = "arch-transaction-handler-$chainId", isDaemon = true) {
+        startArchWorker()
+        startBitcoinWorker()
+    }
+
+    private fun startArchWorker() {
+        archWorkerThread = thread(start = false, name = "arch-transaction-handler-$chainId", isDaemon = true) {
             logger.debug { "Arch Transaction handler thread starting" }
             transaction {
                 programAccount = ArchAccountEntity.findProgramAccount()!!
```
antonyshchenko, line 82: This may throw and without `uncaughtExceptionHandler` I've added in this PR the thread would silently die and we won't get any notifications

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -132,9 +137,17 @@ class ArchTransactionHandler(
                     Thread.sleep(failurePollingInterval.inWholeMilliseconds)
                 }
             }
+        }.also { thread ->
+            thread.uncaughtExceptionHandler = Thread.UncaughtExceptionHandler { _, throwable ->
+                logger.error(throwable) { "Uncaught exception in ${thread.name} thread" }
+                startArchWorker()
```
antonyshchenko, line 143: If there was an unhandled exception in the thread, we log it as error and restart the thread in hope that the error is transient

bholzman: is it worth adding a small delay here before restarting?

bholzman: is it worth adding a small delay here before restarting?

bholzman (approved):
Looks good!

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
antonyshchenko: good point, added

PR #524, opened 2024-09-27
Streamline namings in EVM-related code

antonyshchenko:
Mostly changing `Blockchain` prefix to `Evm` prefix in types and var names plus some package reorg.


backend/src/main/kotlin/xyz/funkybit/core/blockchain/evm/EvmClient.kt
```diff
@@ -53,13 +54,33 @@ import xyz.funkybit.core.utils.toHexBytes
 import java.math.BigDecimal
 import java.math.BigInteger
 import kotlin.jvm.optionals.getOrNull
-
+import kotlin.time.Duration
+
+data class EvmClientConfig(
```
bflood366, line 59: most of this isn't really config associated with the  client but just general EVM configuration for other parts of the system - the only thing in here that the client needs is the URL and maybe the faucet private key but the rest of this is used by other parts of the system.

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
```diff
@@ -3,16 +3,15 @@ package xyz.funkybit.integrationtests.bitcoin
...
 import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.transactions.transaction
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Assumptions
 import org.junit.jupiter.api.extension.ExtendWith
-import xyz.funkybit.core.blockchain.ContractType
 import xyz.funkybit.core.blockchain.bitcoin.ArchNetworkClient
 import xyz.funkybit.core.blockchain.bitcoin.BitcoinClient
+import xyz.funkybit.core.blockchain.evm.ContractType
```
bflood366, line 14: don't think ContractType is really an evm thing - it is chain agnostic. We have the exchange contract on EVM and non EVM.

bflood366 (approved):
looks great - nice job

backend/src/main/kotlin/xyz/funkybit/core/blockchain/evm/EvmClient.kt
antonyshchenko: leaving as is for now, will create a ticket to refactor

integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/bitcoin/ArchOnboardingTest.kt
antonyshchenko: good point, moved to `xyz.funkybit.core.blockchain.ContractType `

backend/src/main/kotlin/xyz/funkybit/core/blockchain/evm/EvmClient.kt
antonyshchenko: Created https://funkybit.atlassian.net/browse/CHAIN-542

PR #525, opened 2024-09-27
Switch to mempool space on backend

bflood366:


.github/workflows/build-push-backend.yml
```diff
@@ -33,7 +33,8 @@ jobs:
         env:
           CI_RUN: 1
           CRYPTOCOMPARE_API_KEY: ${{ secrets.CRYPTOCOMPARE_API_KEY }}
-          BITCOIN_NETWORK_ENABLED: "false"
+          BITCOIN_NETWORK_ENABLED: "true"
+          WAIT_FOR_ARCH_DEPLOYMENT: "true"
```
bflood366, line 37: I have enabled the arch integrations on local CI run - what this env variable does is force it to wait for the arch contract deployment for up to 90 seconds before starting the tests (as oppose to allowing it run in parallel with the tests). If it doesn't start within 90 seconds, it continues on and will run other tests but arch specific tests may fail

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
```diff
@@ -62,17 +60,12 @@ class BitcoinDepositHandler(
     }
 
     private fun refreshPendingDeposit(pendingDeposit: DepositEntity) {
-        val tx = BitcoinClient.getRawTransaction(pendingDeposit.transactionHash)
+        val tx = MempoolSpaceClient.getTransaction(pendingDeposit.transactionHash)
+        if (tx != null && tx.status.confirmed) {
```
bflood366, line 64: this will mark as confirmed now after first confirm.

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
```diff
@@ -81,18 +74,16 @@ class BitcoinDepositHandler(
...
 
-    private fun onChainDepositAmount(tx: BitcoinRpc.Transaction, from: Address): BigInteger? {
+    private fun onChainDepositAmount(tx: MempoolSpaceApi.Transaction): BigInteger {
         return tx
-            .outputsMatchingWallets(setOf(DeployedSmartContractEntity.programBitcoinAddress().value))
-            .find { BitcoinClient.getSourceAddress(tx) == from }
-            ?.value
-            ?.toSatoshi()
-            ?.let(BigInteger::valueOf)
+            .outputsMatchingWallet(DeployedSmartContractEntity.programBitcoinAddress())
```
bflood366, line 86: not quite grokking what this logic was doing before, but now it just sums the outputs of this tx to the program and uses that as the amount of the order.  We shouldn't use the source address since for hierarchical wallets I think the source could be different hierarchical addresses which may not be from the address registered with us.

backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
```diff
@@ -106,6 +106,7 @@ sealed class DefaultBlockParam {
 
 open class BlockchainClient(val config: BlockchainClientConfig) {
     protected val web3jService: HttpService = httpService(config.url, config.enableWeb3jLogging)
+    protected val web3jServiceNoLogging: HttpService = httpService(config.url, logging = false)
```
bflood366, line 109: might be a better way to do this - but I had the anvil_mine use this HttpService so test run logs are not polluted with anvil_mine logs

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ProgramUtxoRefresherTask.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import org.jetbrains.exposed.sql.transactions.transaction
+import xyz.funkybit.core.blockchain.bitcoin.bitcoinConfig
+import xyz.funkybit.core.model.BitcoinAddress
+import xyz.funkybit.core.model.db.DeployedSmartContractEntity
+import xyz.funkybit.core.services.UtxoManager
+import kotlin.time.Duration.Companion.seconds
+
+class ProgramUtxoRefresherTask : RepeaterBaseTask(
```
bflood366, line 11: this task will be called every 30 seconds and also triggered from the arch/bitcoin handlers when deposits or withdrawals are confirmed or if we try to do a withdrawal and don't have enough assets

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
antonyshchenko: I've added that logic in scope of this PR https://github.com/funkybit-labs/chainring-monorepo/pull/520 
The idea is that we should not just blindly trust the amount provided by the user when they notify us about deposit via API

backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
antonyshchenko: I'm wondering if in general we should keep logging all HTTP calls our system makes. Is that useful? Could we perhaps log only certain calls when it matters? We could for example shift the responsibility of deciding which calls should be logged to the user of `BlockchainClient`.

antonyshchenko: Not proposing to change this in this PR, just thinking about how we could reduce the noise in the logs

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
bflood366: I do think the logic may not have have always worked since it was trying to match the source address which might not alway work with hierarchical wallets - this just now sums the outputs to the program address (should only ever be 1) and uses that amount to match and update if different.

backend/src/main/kotlin/xyz/funkybit/core/blockchain/BlockchainClient.kt
bflood366: yeah seems like moving the decision to log or not log up a level would be better or maybe we somehow need more finer grained controlled so we can turn certain components on or off. I guess we could do that with log4j config. There is so much noise in the logs when testing at least it makes it difficult to track down things some times.

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
antonyshchenko: That's fine, however there might be a room for misuse still. I don't know if this is practical, but someone could watch mempool, find other people's transactions with transfer to our program address and try submitting them as if it was their own. Not sure if we should worry about this right now, but might be worth creating a ticket.

bflood366: It definitely would be possible if someone sent a tx to our program, but then never called our API endpoint, someone could take that tx and call our endpoint. Ideally for hierarchical wallet if they could somehow send us the xpub key then we could derive some finite number of possible addresses and check against that. WE should maybe log on a successful bitcoin deposit the wallet address for the deposit and the source addresses of the input utxos - then we could at least have a data point as to how common this is. I'll add a ticket so we can look into how to solve this.

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
bflood366: we said for testnet and mainnet we only want 1 confirm since blocks are approx every 10 minutes.  So as of now there is no env we would ever want it to be more than 1. Easy to add back if needed

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -374,9 +378,12 @@ class ArchTransactionHandler(
                             tx = inProgressBatch.preparationTx,
                         ) { _, error ->
                             if (error == null) {
-                                val onChainBatchHash = getOnChainBatchHash()
+                                val onChainBatchHash = getOnChainBatchHash(retryIfEmpty = true)
                                 if (onChainBatchHash != inProgressBatch.preparationTx.batchHash) {
-                                    inProgressBatch.markAsFailed("Batch hash mismatch for settlement batch ${inProgressBatch.guid.value}, on chain value: $onChainBatchHash, db value: ${inProgressBatch.preparationTx.batchHash}")
+                                    "Batch hash mismatch for settlement batch ${inProgressBatch.guid.value}, on chain value: $onChainBatchHash, db value: ${inProgressBatch.preparationTx.batchHash}".also {
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -374,9 +378,12 @@ class ArchTransactionHandler(
...
+                                        logger.debug { it }
```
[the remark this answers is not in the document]
bflood366: yeah it should have been a `logger.error` - changes

backend/src/main/kotlin/xyz/funkybit/apps/ring/ArchTransactionHandler.kt
```diff
@@ -534,12 +558,16 @@ class ArchTransactionHandler(
...
             submitToArch(transaction, instruction)
-            UtxoSelectionService.reserveUtxos(selectedUtxos, transaction.txHash?.value ?: "")
+            UtxoManager.reserveUtxos(selectedUtxos, transaction.txHash?.value ?: "")
 
             return true
         }
     }
 
+    private fun initiateProgramUtxoRefresh() {
+        TransactionManager.current().notifyDbListener(REPEATER_APP_TASK_CTL_CHANNEL, "program_utxo_refresher")
```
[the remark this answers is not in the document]
bflood366: added a util function as suggested

backend/src/main/kotlin/xyz/funkybit/core/blockchain/bitcoin/BitcoinClient.kt
```diff
@@ -3,281 +3,85 @@ package xyz.funkybit.core.blockchain.bitcoin
...
+        }
 
-    fun sendToAddress(address: BitcoinAddress, amount: BigInteger): TxHash {
-        return getValue(
-            BitcoinRpcRequest(
-                "sendtoaddress",
-                BitcoinRpcParams(listOf(address.value, amount.inSatsAsDecimalString())),
-            ),
-        )
+        miningThread = thread(start = true, name = "bitcoin mining", isDaemon = true) {
```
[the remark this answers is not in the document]
bflood366: moved out the mining peice

backend/src/main/kotlin/xyz/funkybit/core/model/db/Wallet.kt
```diff
@@ -93,6 +93,19 @@ class WalletEntity(guid: EntityID<WalletId>) : GUIDEntity<WalletId>(guid) {
...
 
+        fun findByAddresses(addresses: List<Address>): WalletEntity? {
+            return WalletEntity.find {
+                WalletTable.address.inList(
+                    addresses.map { address ->
+                        when (address) {
+                            is EvmAddress -> address.canonicalize().value
+                            is BitcoinAddress -> address.canonicalize().value
+                        }
+                    },
```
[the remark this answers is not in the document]
bflood366: changed

backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
bflood366: ok, for now I added back logic to check against num confirms but will be 1 for now - sounds like the logic will need to be conditioned on other criteria and not just a single number.

backend/src/main/kotlin/xyz/funkybit/core/blockchain/bitcoin/MempoolSpaceClient.kt
```diff
@@ -0,0 +1,198 @@
...
+    }
+    val logger = KotlinLogging.logger {}
+
+    private val mediaType = "text/plain".toMediaTypeOrNull()
+
+    private val minFee = bitcoinConfig.feeSettings.minValue.toBigInteger()
+    private val maxFee = bitcoinConfig.feeSettings.maxValue.toBigInteger()
+
+    private fun execute(request: Request): Response {
+        logger.debug { "Request -> ${request.method} ${request.url}" }
```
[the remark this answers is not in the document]
bflood366: I am not going to change anything yet on logging since it sounds like you are looking into reducing that noise and what ever pattern we use for the clients to reduce the noice thy should all follow that pattern

PR #526, opened 2024-09-30
Periodic cleanup of broadcaster_jobs table

antonyshchenko:
On demo env this table has accumulated 29 GB of data within 20 days

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/model/db/BroadcasterJob.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/model/db/BroadcasterJob.kt
@@ -3,10 +3,13 @@ package xyz.funkybit.core.model.db
 import de.fxlae.typeid.TypeId
 import io.github.oshai.kotlinlogging.KotlinLogging
 import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
 import kotlinx.serialization.Serializable
 import org.http4k.format.KotlinxSerialization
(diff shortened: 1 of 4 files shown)


backend/src/main/kotlin/xyz/funkybit/core/model/db/BroadcasterJob.kt
```diff
@@ -65,13 +68,16 @@ object BroadcasterJobTable : GUIDTable<BroadcasterJobId>("broadcaster_job", ::Br
...
+        fun create(id: BroadcasterJobId, notifications: List<BroadcasterNotification>, currentTime: Instant): BroadcasterJobId {
+            return BroadcasterJobEntity.new(id) {
+                this.createdAt = currentTime
                 this.createdBy = "system"
                 this.notificationData = notifications
             }.guid.value
         }
+
+        fun deleteOlderThan(timestamp: Instant): Int =
+            BroadcasterJobTable.deleteWhere { createdAt.less(timestamp) }
```
bholzman, line 80: there's no index on createdAt so this will require a table scan

bholzman (approved):

backend/src/main/kotlin/xyz/funkybit/core/model/db/BroadcasterJob.kt
antonyshchenko: Good point, added an index

PR #528, opened 2024-09-30
fix for sporadic deposit failure in arch on CI

bflood366:
the primary issue was for arch we have an extra state since we need to send a tx to arch after we detect the deposits on chain, but that was not represented by a separate state in the model - it just used `Settling` - this adds a different state to represent that state to represent when it has been `SentToArch` and it only goes to `Settling` after arch has processed the tx. The lack of a separate state led to some sporadic failures since we were updating tables in the arch transaction handler and calling the sequencer in the same transaction and sometimes the sequencer response processor would run and complete it's tx, before the arch transaction handler completed it's tx which led to some sporadic test failures.


antonyshchenko (approved):
Looks great! 👍

PR #529, opened 2024-10-01
demo env rds instance with regular postgres engine

imykolenko:


imykolenko:
Closing this PR as identified root cause is not related to db engine type.

PR #531, opened 2024-10-02
Bitcoin auto-mining when running locally, bump contracts to bring in changes required to allow testin with Xverse wallet locally

antonyshchenko:


terraform/testnet/main.tf
```diff
@@ -107,14 +107,14 @@ module "waf" {
   aws_region = var.aws_region
   aws_lb_arn = module.alb.lb_arn
   #aws_lb_arn               = module.alb.https_listener_arn
-  short_env_name           = local.name_prefix
-  ipv4_blacklist_addresses = local.ipv4blacklist
-  ipv4_whitelist_addresses = local.ipv4whitelist
-  ipv6_blacklist_addresses = local.ipv6blacklist
-  ipv6_whitelist_addresses = local.ipv6whitelist
-  blocked_country_codes    = var.blocked_country_codes
+  short_env_name                = local.name_prefix
```
antonyshchenko, line 110: Result of `make format`

src/main/kotlin/xyz/funkybit/LocalMain.kt
```diff
@@ -30,6 +31,7 @@ fun main(args: Array<String>) {
                 RingApp().start()
                 ApiApp().start()
                 BotApp().start()
+                BitcoinMiner.start()
```
antonyshchenko, line 34: Without this Bitcoin deposits and settlements get stuck locally

backend/src/test/kotlin/xyz/funkybit/core/bitcoin/UtxoManagerTest.kt
```diff
@@ -311,7 +311,7 @@ class UtxoManagerTest : TestWithDb() {
             ),
         )
 
-        every { MempoolSpaceClient.getTransactions(feePayerAddress, txId4) } returns listOf()
+        every { MempoolSpaceClient.getTransactions(feePayerAddress, txId1) } returns listOf()
```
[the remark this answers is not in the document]
bflood366: ah good catch

terraform/testnet/main.tf
bflood366: I think we had some terraform for saving fulcrum image in ECR - can be removed.

bflood366 (approved):
looks great

terraform/testnet/main.tf
antonyshchenko: yup, done

PR #535, opened 2024-10-04
backend changes for arch program/token account layouts

bflood366:


backend/src/main/kotlin/xyz/funkybit/core/model/TxHash.kt
```diff
@@ -21,6 +21,10 @@ value class TxHash(val value: String) {
             return TxHash(ByteArray(32).toHex())
         }
 
+        fun emptyHashArch(): TxHash {
```
antonyshchenko, line 24: This and `emptyHash()` method above could actually be defined as vals

backend/src/main/kotlin/xyz/funkybit/core/model/bitcoin/ProgramInstruction.kt
```diff
@@ -121,6 +124,15 @@ enum class BitcoinNetworkType {
                 else -> Signet
             }
         }
+
+        fun fromByte(ordinal: Int): BitcoinNetworkType {
+            return when (ordinal) {
+                0 -> Bitcoin
```
antonyshchenko, line 130: Nit: Why `Bitcoin` instead of `Mainnet`? It might be better to stick to domain terminology.

backend/src/main/kotlin/xyz/funkybit/core/model/bitcoin/ProgramInstruction.kt
```diff
@@ -197,28 +209,129 @@ object ProgramInstructionSerializer : KSerializer<ProgramInstruction> {
...
+        @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
+        override val descriptor: SerialDescriptor = buildSerialDescriptor("Program", StructureKind.LIST)
+
+        override fun serialize(encoder: Encoder, value: Program) {
+            throw Exception("not required")
+        }
+
+        override fun deserialize(decoder: Decoder): Program {
+            return when (decoder) {
+                is com.funkatronics.kborsh.BorshDecoder -> {
```
antonyshchenko, line 249: I thought we are removing `borsh` serialization. Or are we using a mixture of both `borsh` and handrolled serialization?

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/TxHash.kt
bflood366: changed these to vals

backend/src/main/kotlin/xyz/funkybit/core/model/bitcoin/ProgramInstruction.kt
bflood366: I had initially introduced this is in the program and in rust bitcoin crate they use Bitcoin, Testnet and Regtest as the network type enum names, but java bitcoinj uses Mainnet, Testnet and Regtest. I changed this to Mainnet to be consistent on kotlin side

backend/src/main/kotlin/xyz/funkybit/core/model/bitcoin/ProgramInstruction.kt
bflood366: WE have not removed borsh entirely - the instructions still use borsh on the program/kotlin side - we have a ticket to remove borsh there too.  Here for the account state serialization we  just using the convenience of borsh being hooked into the kotlinx serialization since we are doing our own custom serializers. WE could probably remove borsh here too and write out own decoder/encoder for the kotlinx serialization.

PR #539, opened 2024-10-08
CHAIN-525 show correct base symbol

imykolenko:
currently both buy/sell show same value for buy orders


bholzman (approved):
Whoops, thanks for fixing

PR #541, opened 2024-10-11
Optimize recent orders and trades fetching

antonyshchenko:
Order table in demo has around 70M records and fetching recent orders for market maker takes minutes affecting overall database performance quite badly.

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
@@ -44,7 +44,7 @@ object OrderExecutionTable : GUIDTable<ExecutionId>("order_execution", ::Executi
     ).index()
     val feeAmount = decimal("fee_amount", 30, 0)
     val feeSymbol = varchar("fee_symbol", 10485760)
-    val marketGuid = reference("market_guid", MarketTable).nullable()
+    val marketGuid = reference("market_guid", MarketTable)
     val side = customEnumeration(
         "side",
         "OrderSide",
@@ -52,6 +52,18 @@ object OrderExecutionTable : GUIDTable<ExecutionId>("order_execution", ::Executi
         { PGEnum("OrderSide", it) },
     )
     val userGuid = reference("user_guid", UserTable).index()
+    val responseSequence = long("response_sequence").index()
+
+    init {
+        OrderTable.index(
+            customIndexName = "order_execution_user_guid_timestamp_index",
+            columns = arrayOf(userGuid, timestamp),
+        )
+        OrderTable.index(
+            customIndexName = "order_execution_user_guid_response_sequence_index",
+            columns = arrayOf(userGuid, responseSequence),
+        )
+    }
 }
 
 class OrderExecutionEntity(guid: EntityID<ExecutionId>) : GUIDEntity<ExecutionId>(guid) {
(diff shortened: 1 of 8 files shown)


backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V98_AddUserGuidToOrderAndExecution.kt
```diff
@@ -0,0 +1,34 @@
...
+import xyz.funkybit.core.db.Migration
+
+@Suppress("ClassName")
+class V98_AddUserGuidToOrderAndExecution : Migration() {
+    override fun run() {
+        transaction {
+            exec(
+                """
+                ALTER TABLE "order" ADD COLUMN user_guid CHARACTER VARYING(10485760);
+                UPDATE "order" SET user_guid = (
```
antonyshchenko, line 13: This takes around 30 minutes locally on demo db snapshot so I propose to reset the demo env.

backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
```diff
@@ -118,46 +120,36 @@ class OrderExecutionEntity(guid: EntityID<ExecutionId>) : GUIDEntity<ExecutionId
...
 
         fun listForUser(userId: EntityID<UserId>, beforeTimestamp: Instant, limit: Int): List<OrderExecutionEntity> {
             return OrderExecutionTable
-                .join(OrderTable, JoinType.INNER, OrderTable.guid, OrderExecutionTable.orderGuid)
-                .innerJoin(WalletTable)
                 .join(TradeTable, JoinType.INNER, TradeTable.guid, OrderExecutionTable.tradeGuid)
                 .select(OrderExecutionTable.columns)
                 .where {
-                    WalletTable.userGuid.eq(userId) and OrderExecutionTable.timestamp.less(beforeTimestamp)
+                    OrderExecutionTable.userGuid.eq(userId) and OrderExecutionTable.timestamp.less(beforeTimestamp)
```
imykolenko, line 148: it looks like composite index by `userId` and `timestamp` could be useful here

backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
```diff
@@ -118,46 +120,36 @@ class OrderExecutionEntity(guid: EntityID<ExecutionId>) : GUIDEntity<ExecutionId
...
-                .innerJoin(WalletTable)
                 .select(TradeTable.responseSequence)
                 .withDistinct(true)
                 .where {
-                    WalletTable.userGuid.eq(userId)
+                    OrderExecutionTable.userGuid.eq(userId)
                 }
                 .orderBy(TradeTable.responseSequence, SortOrder.DESC)
                 .limit(maxSequencerResponses)
                 .mapNotNull { it[TradeTable.responseSequence] }
```
imykolenko, line 130: This query filters and sorts by different tables. Coming to a result means either scaning backwards by response sequence and collect limit records with matching userID OR build a virtual list of all executions for a given user and then sort it by sequence number and then take limit. Both can take a while.

To solve this I'd propose to store `responseSequence` also on OrderExecution, then it becomes a query on one single table/index.

backend/src/main/kotlin/xyz/funkybit/core/model/db/migrations/V98_AddUserGuidToOrderAndExecution.kt
imykolenko: demo reset is rest, should be quick now :)

imykolenko (approved):
looks great, nice work optimizing data access speed!

backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
antonyshchenko: added

backend/src/main/kotlin/xyz/funkybit/core/model/db/OrderExecution.kt
antonyshchenko: done

PR #543, opened 2024-10-14
Demo reset instructions update

antonyshchenko:


imykolenko (approved):
Nice!

Btw, what if we switch to complete db cleanup not preserving OHKC records? I know I did it time ago, but now I think it is better to have completely new start. 

`drop owned by funkybit` just clears the db without need to recreate.

PR #545, opened 2024-10-14
Chain 407/api rate limiting

imykolenko:
* friendly WAF response (429 -> `{"message":"Request rate limit exceeded. Please try again later."}`)
* UI message in the same style as maintenance
* interceptor that processed response code is also added to `noAuthApiClient` that is used to fetch configuration to make sure maintenance mode is recognized on page reload

===
Screenshot 2024-10-14 at 2 28 32 PM

### DIFF
--- a/web-ui/src/apiClient.ts
+++ b/web-ui/src/apiClient.ts
@@ -1018,16 +1018,25 @@ export const noAuthApiClient = new Zodios(apiBaseUrl, [
   }
 ])
 
-export function useAccessRestriction() {
-  const [maintenance, setMaintenance] = useState(false)
-  const [restricted, setRestricted] = useState(false)
-  const [rateLimited, setRateLimited] = useState(false)
+export type AccessRestriction =
+  | 'none'
+  | 'underMaintenance'
+  | 'geoBlocked'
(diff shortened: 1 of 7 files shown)


web-ui/src/apiClient.ts
```diff
@@ -1020,34 +1021,35 @@ export const noAuthApiClient = new Zodios(apiBaseUrl, [
...
-          if (maintenance) {
-            setMaintenance(false)
-          }
-          if (restricted) {
-            setRestricted(false)
-          }
+    const handleResponseCode = (status: number) => {
+      setMaintenance(status === 418)
+      setRestricted(status === 451)
+      setRateLimited(status === 429)
```
antonyshchenko, line 1030: Seems like these states are mutually-exclusive. How about turning it into some kind of enum? E.g `'maintenance' | 'restricted' | 'rateLimited'`

antonyshchenko (approved):
Looks great!

web-ui/src/apiClient.ts
imykolenko: yes, they are. Updated.

PR #546, opened 2024-10-15
CHAIN-557 extend low gas alert

imykolenko:
extend gas monitor to cover testnet challenge airdropoer gas/usdc and bitcoin fee payer


backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/BalancesMonitorTask.kt
```diff
@@ -0,0 +1,145 @@
...
+                    }
+                }
+            }
+        } finally {
+            nextInvocationArgs = null
+            HttpClient.setQuietModeForThread(false)
+        }
+    }
+
+    private fun checkBitcoinFeePayerGas(minimumGas: BigInteger = BigInteger.valueOf(100_000)) {
```
bholzman, line 72: should these defaults be in the config? I'm also thinking we might want a bit of a higher minimum here, maybe more like 0.1 or 0.05 btc, and maybe more like 500 or 1000 minEnrollments for testnet challenge

bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/BalancesMonitorTask.kt
```diff
@@ -0,0 +1,145 @@
...
+import xyz.funkybit.core.model.db.SymbolEntity
+import xyz.funkybit.core.services.UtxoManager
+import xyz.funkybit.core.utils.HttpClient
+import xyz.funkybit.core.utils.TestnetChallengeUtils
+import xyz.funkybit.core.utils.toFundamentalUnits
+import java.math.BigInteger
+import kotlin.time.Duration.Companion.minutes
+
+data class BalancesMonitorConfig(
+    val testnetChallengeEnabled: Boolean = System.getenv("TESTNET_CHALLENGE_ENABLED")?.toBoolean() ?: true,
```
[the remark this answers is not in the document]
imykolenko: thx!

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/BalancesMonitorTask.kt
imykolenko: good point! Moved to config, bumped defaults to 0.1 btc and 1000 min enrollements.

PR #550, opened 2024-10-16
CHAIN-563: disable telegram bot for now

bholzman:


backend/src/main/resources/exchangeprogram.so
bflood366, line 1: hmm why did this change

bflood366 (approved):
looks good

PR #551, opened 2024-10-17
Chain 578/integrate sentry for tracing

imykolenko:
- tracing is implemented using Opentelemetry, Sentry is only used to store traces. This should allow us to easily change service we use (newrelic/honeycomb/selfhosted)
 - grpc interceptor is standard from opentelemetry
 - okHttp interceptor is custom, but only to reduce amount of data. Only requests that are performed in the context of existing trace are recorded
  - same for sql queries


bholzman (approved):
Looks good!

PR #564, opened 2024-10-21
Attempt to fix occasional checkpoint skipping in Sequencer

antonyshchenko:


sequencer/src/test/kotlin/xyz/funkybit/TestSequencerCheckpoints.kt
```diff
@@ -395,7 +395,7 @@ class TestSequencerCheckpoints {
             }
 
             assertQueueFilesCount(inputQueue, 2)
-            assertCheckpointsCount(checkpointsQueue, 1)
+            assertCheckpointsCount(checkpointsQueue, 2)
```
antonyshchenko, line 398: +1 here because it will produce an empty checkpoint in the very beginning when it starts processing the first cycle

imykolenko (approved):

PR #569, opened 2024-10-23
X (Twitter) account linking in Testnet

antonyshchenko:
The card in UI says "Link your Discord/X account to earn more points!" however I could not find where do we actually grant points for this. Is it out of the scope for now and do we have a ticket to address that?

### DIFF
--- a/backend/src/main/kotlin/xyz/funkybit/core/model/db/UserLinkedAccount.kt
+++ b/backend/src/main/kotlin/xyz/funkybit/core/model/db/UserLinkedAccount.kt
@@ -1,0 +1,118 @@
+package xyz.funkybit.core.model.db
+
+import kotlinx.datetime.Clock
+import kotlinx.serialization.Serializable
+import org.jetbrains.exposed.dao.IntEntity
+import org.jetbrains.exposed.dao.IntEntityClass
+import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.kotlin.datetime.timestamp
+import org.jetbrains.exposed.sql.statements.UpsertStatement
+import org.jetbrains.exposed.sql.stringLiteral
+import org.jetbrains.exposed.sql.upsert
+import xyz.funkybit.core.utils.OAuth2
+
+@Serializable
+enum class UserLinkedAccountType {
+    Discord,
+    X,
+}
+
+object UserAccountLinkingIntentTable : IntIdTable("user_account_linking_intent") {
+    val userGuid = reference("user_guid", UserTable)
+    val type = customEnumeration(
+        "type",
+        "UserLinkedAccountType",
+        { value -> UserLinkedAccountType.valueOf(value as String) },
+        { PGEnum("UserLinkedAccountType", it) },
+    )
+    val oauth2CodeVerifier = varchar("oauth2_code_verifier", 10485760).nullable()
+    val createdAt = timestamp("created_at")
+
+    init {
+        index(true, userGuid, type)
+    }
+}
+
+class UserAccountLinkingIntentEntity(id: EntityID<Int>) : IntEntity(id) {
+    companion object : IntEntityClass<UserAccountLinkingIntentEntity>(UserAccountLinkingIntentTable) {
+        fun create(user: UserEntity, accountType: UserLinkedAccountType) {
+            UserAccountLinkingIntentTable.upsert(
+                keys = arrayOf(UserAccountLinkingIntentTable.userGuid, UserAccountLinkingIntentTable.type),
+                onUpdate = listOfNotNull(
+                    if (accountType == UserLinkedAccountType.X) {
+                        UserAccountLinkingIntentTable.oauth2CodeVerifier to stringLiteral(OAuth2.CodeVerifier.generate().value)
+                    } else {
+                        null
+                    },
+                    UserAccountLinkingIntentTable.createdAt to stringLiteral("now()"),
+                ),
+                body = fun UserAccountLinkingIntentTable.(it: UpsertStatement<Long>) {
+                    it[this.userGuid] = user.guid
+                    it[this.type] = accountType
+                    if (accountType == UserLinkedAccountType.X) {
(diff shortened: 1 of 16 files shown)


backend/src/main/kotlin/xyz/funkybit/core/utils/XClient.kt
```diff
@@ -0,0 +1,89 @@
...
+            .query("redirect_uri", redirectUri)
+            .query("scope", listOf("offline.access", "users.read", "tweet.read").joinToString(" "))
+            .query("response_type", "code")
+            .query("code_challenge", codeChallenge.value)
+            .query("code_challenge_method", "S256")
+            .query("state", UUID.randomUUID().toString().replace("-", ""))
+            .toString()
+
+    fun getAuthTokens(authorizationCode: OAuth2.AuthorizationCode, codeVerifier: OAuth2.CodeVerifier): OAuth2.Tokens {
+        val credentials = "$clientId:$clientSecret".base64Encode()
```
antonyshchenko, line 47: Took me a while to figure out how to make this call properly. X documentation is a mess...

bholzman:
Good catch, I had meant to add a ticket for granting points -- added https://funkybit.atlassian.net/browse/CHAIN-596

backend/src/main/kotlin/xyz/funkybit/core/model/db/UserLinkedAccount.kt
```diff
@@ -0,0 +1,118 @@
...
+    val oauth2AccessToken = varchar("oauth2_access_token", 10485760)
+    val oauth2RefreshToken = varchar("oauth2_refresh_token", 10485760)
+    val createdAt = timestamp("created_at")
+
+    init {
+        index(true, userGuid, type)
+    }
+}
+
+class UserLinkedAccountEntity(id: EntityID<Int>) : IntEntity(id) {
```
bholzman, line 98: as you may notice in my telegram discord linking PR, we also need to link to accounts from TelegramMiniAppUsers ... a few different options I guess -- 
1) could remove the user from this table and add separate linking tables from users to linked account and from tma users to linked account
2) could duplicate this to TelegramMiniAppUserLinkedAccountEntity
3) could always create an underlying User entity for every TelegramMiniAppUser entity.

What do you think makes the most sense?

backend/src/main/kotlin/xyz/funkybit/core/utils/OAuth2.kt
```diff
@@ -0,0 +1,48 @@
...
+import kotlinx.serialization.Serializable
+import org.http4k.base64Encode
+import java.util.UUID
+
+object OAuth2 {
+    @JvmInline
+    value class CodeVerifier(val value: String) {
+        companion object {
+            fun generate(): CodeVerifier =
+                CodeVerifier(UUID.randomUUID().toString().replace("-", ""))
```
bholzman, line 13: does this need cryptographic randomness (i.e. SecureRandom)?

backend/src/main/kotlin/xyz/funkybit/core/utils/XClient.kt
```diff
@@ -0,0 +1,89 @@
...
+import okhttp3.FormBody
+import okhttp3.Request
+import org.http4k.base64Encode
+import org.http4k.core.Uri
+import org.http4k.core.query
+import java.util.UUID
+
+object XClient {
+    private val clientId = System.getenv("X_CLIENT_ID") ?: ""
+    private val clientSecret = System.getenv("X_CLIENT_SECRET") ?: "-Hpxck"
```
bholzman, line 16: I assume this default is not actually a valid client secret -- any reason the default is not just `""` ?

backend/src/main/kotlin/xyz/funkybit/core/utils/XClient.kt
bholzman: why does that not surprise me...

web-ui/.env.testnet
```diff
@@ -1,4 +1,3 @@
 ENV_API_URL=https://testnet-api.funkybit.fun
 ENV_WEB_URL=https://testnet.funkybit.fun
 ENV_TESTNET_CHALLENGE_ENABLED=true
-ENV_DISCORD_CLIENT_ID=1296545058736050297
```
bholzman, line 4: My telegram discord linking PR uses this env variable too -- the oauth2 url for that flow is slightly different than for the webapp flow (since we need to request the guilds.join scope there)

bholzman (approved):

backend/src/main/kotlin/xyz/funkybit/core/model/db/UserLinkedAccount.kt
antonyshchenko: I like option 3 the most, tbh

backend/src/main/kotlin/xyz/funkybit/core/utils/OAuth2.kt
antonyshchenko: You are right, the spec says it should be cryptographically random. Changed

backend/src/main/kotlin/xyz/funkybit/core/utils/XClient.kt
antonyshchenko: that's a leftover, removed

web-ui/.env.testnet
antonyshchenko: I've unified the flow for Discord and X in this PR and now the authorize url is returned by the backend. We might to do the same for oauth2-relay. `/v1/oauth-relay-auth-token` endpoint could return the authorize url alongside the relay token. This way we won't need Discord/X client ids in UI. What do you think?

backend/src/main/kotlin/xyz/funkybit/core/model/db/UserLinkedAccount.kt
bholzman: one possible downside I see to that is we are adding additional joins to retrieve the linked account data, which we will need every time every tma user gets their state, so if we end up getting lots of telegram users it might be a bit resource intensive. but we can always denormalize if and when that happens

web-ui/.env.testnet
antonyshchenko: Ok, I've brought this constant back here

PR #570, opened 2024-10-23
CHAIN-546: Allow TG users to link their discord account

bholzman:


telegram-mini-app/.env.testnet
```diff
@@ -1,3 +1,4 @@
 ENV_API_URL=https://testnet-api.funkybit.fun
+ENV_API_URL=https://testnet.funkybit.fun
```
imykolenko, line 2: Looks like should be `ENV_WEB_URL`

imykolenko (approved):
Nice work, looks great!

backend/src/main/kotlin/xyz/funkybit/apps/api/TelegramMiniAppRoutes.kt
```diff
@@ -254,4 +262,57 @@ object TelegramMiniAppRoutes {
             }
         }
     }
+
+    val oauthRelayAuthToken: ContractRoute = run {
+        val responseBody = Body.auto<OauthRelayToken>().toLens()
+
+        "oauth-relay-auth-token" meta {
```
antonyshchenko, line 269: This endpoint could also create `UserAccountLinkingIntentEntity` and return the authorization url, but it would have to take `UserLinkedAccountType` as param. Or perhaps the better way to do it would to change the front-end part of oauth-relay to call the `startAccountLinking`/`completeAccountLinking` endpoints instead of `completeDiscordLinking` here. Above-mentioned endpoints would have to be aware that this is a TMA flow though.

bholzman: the completeDiscordLinking call is secured differently than the completeAccountLinking endpoint though...

telegram-mini-app/.env.testnet
bholzman: thanks!

PR #573, opened 2024-10-24
restore sequencer state from current cycle's checkpoint

imykolenko:
The checkpoint trigger has been changed from relying on tailerState == TailerState.END_OF_CYCLE to detecting a change in the inputTailer's cycle. As a result, the checkpoint is now saved at the start of a new cycle instead of at the end of the current one. 
Restoration logic has been updated accordingly.


antonyshchenko (approved):
Nice find! 👍

PR #574, opened 2024-10-24
restore from last existing checkpoint instead of searching going backward

imykolenko:
Input tailor was moved almost to the beginning which block checkpoint restoration due to input queue cycle and checkpoint cycle mismatch.
Changed restoration to always take last available checkpoint.


antonyshchenko (approved):
👍

PR #579, opened 2024-10-25
CHAIN-559: persist BTC network fee for withdrawals

bflood366:


backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/ArchUtils.kt
```diff
@@ -453,17 +453,26 @@ object ArchUtils {
...
+                    TransactionOutPoint(
+                        params,
+                        it.vout(),
+                        Sha256Hash.wrap(it.txId().value),
+                    ),
+                    Coin.valueOf(it.amount.toLong()),
+                ).also { input ->
+                    input.setScriptSig(ScriptBuilder.createEmpty())
+                    // this matches the size of the elements pushed onto witness stack when arch signs an input
+                    input.witness = TransactionWitness(3).also { transactionWitness ->
```
bflood366, line 470: the way the fee is calculated is we get the recommended fee and then build a tx to emulate what the program will send and multiply the tx vsize by recommended fee (sats/vbyte). We then use that given the input and output utxos to determine the change amount.  I noticed that the vsize of the actual tx sent on chain and our estimate here was different (so the effective sats/vbyte we were paying was lower than the recommended fee). The issue was we using P2WPKH inputs in the estimate, but the inputs are taproot - so this builds a witness stack with same length items as the program builds for the taproot inputs.

backend/src/main/kotlin/xyz/funkybit/core/model/db/BlockchainTransaction.kt
```diff
@@ -82,6 +83,13 @@ class BlockchainTransactionEntity(guid: EntityID<BlockchainTransactionId>) : GUI
             }
             return entity
         }
+
+        fun findByChainAndHash(chainId: ChainId, txHash: TxHash): BlockchainTransactionEntity? {
+            return BlockchainTransactionEntity.find {
+                BlockchainTransactionTable.chainId.eq(chainId) and
```
bholzman, line 89: noticed we don't have an index on either txHash or chainId (probably just on txHash is sufficient)

bholzman (approved):
Looks good!

backend/src/main/kotlin/xyz/funkybit/core/model/db/BlockchainTransaction.kt
bflood366: added an index on tx_hash

PR #584, opened 2024-10-28
limit makers batch frequency

imykolenko:
when maker reacts on every price update it quickly gets to 10 batches per second. Nice that it works, but not informative.


bholzman:
> when maker reacts on every price update it quickly gets to 10 batches per second. Nice that it works, but not informative.

So in testnet I'm thinking we want the maker to react whenever either the external price feed updates or when an execution makes the market price differ from the external price feed by more than a small amount (for some definition of small). Seems like waiting up to 2 seconds to update might make it seem like our price is lagging behind the true market price.

bholzman (approved):
Looks good!

PR #586, opened 2024-10-28
CHAIN-617: add point multiplier to x/discord linking in web app

bholzman:


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
```diff
@@ -240,7 +245,7 @@ fun UserEntity.pointsBalances(): Map<TestnetChallengeUserRewardType, BigDecimal>
         .where { TestnetChallengeUserRewardTable.userGuid.eq(guid) }
         .groupBy(typeColumn)
         .associate {
-            it[typeColumn] to (it[sumColumn]?.setScale(18) ?: BigDecimal.ZERO)
+            it[typeColumn] to (it[sumColumn]?.setScale(18) ?: BigDecimal.ZERO).times(this.testnetPointMultiplier.toBigDecimal())
```
imykolenko, line 248: So multiplier is retrospective and affects all earned points, that might be great motivation booster!

Does multiplier affect referral bonus?

imykolenko (approved):
Looks great

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengeUserReward.kt
bholzman: I had thought about it and decided it was fine not to (and is obviously simpler). Your multiplier is still effectively applied to the referral bonus you receive.

PR #587, opened 2024-10-28
Arch chain id

antonyshchenko:
- Change chain id type to string
- Add chain id for Arch network
- Distinguish Arch transactions from Bitcoin transactions in the database
- ~~Arch Bitcoin token now has `arch` as chain id~~
- Arch chain is not exposed via API
- Decided not to introduce a prefix for EVM chain ids since it would require more changes in UI


backend/src/main/kotlin/xyz/funkybit/apps/api/ConfigRoutes.kt
```diff
@@ -118,33 +118,37 @@ class ConfigRoutes(private val faucetMode: FaucetMode) {
...
-                                    jsonRpcUrl = chain.jsonRpcUrl,
-                                    blockExplorerNetName = chain.blockExplorerNetName,
-                                    blockExplorerUrl = chain.blockExplorerUrl,
-                                    networkType = chain.networkType,
-                                )
-                            },
+                            chains = ChainEntity
+                                .all()
+                                .orderBy(ChainTable.id to SortOrder.ASC)
+                                .filter { it.id.value != ChainId.bitcoin }
```
bflood366, line 124: why is this not Bitcoin - seems like we must be mapping arch to bitcoin then in UI - I would think the UI or anyone using the API would see EVM chains and bitcoin. To some extent arch is an implementation detail to the users of the API since it does not talk directly to it - it talks directly to bitcoin. The address we return for the deployed contract is actually address on bitcoin network to deposit funds to and the block explorers are for the bitcoin network. If we were to switch arch out for something else that worked similarly on bitcoin network I don't think that implementation detail should filter out via the API.

I would think leaving deposits and withdrawals and symbols being associated with bitcoin chain as oppose to arch chain is a more accurate representation of our model.

antonyshchenko: Good point, changed per your suggestion

bflood366 (approved):

bflood366:
looks like there are compilation failures in mocker

PR #588, opened 2024-10-29
Chain 592/limit maker update frequency 2

imykolenko:


mocker/src/main/kotlin/xyz/funkybit/mocker/core/Maker.kt
```diff
@@ -381,7 +385,8 @@ class Maker(
         logger.debug { "$id: creating quotes in market $marketId, current price: $curPrice, price override: $marketPriceOverride" }
         markets.find { it.id == marketId }?.let { market ->
             apiClient.cancelOpenOrders()
-            currentOrders[marketId] = mutableSetOf()
+            // current orders are cancelled in the batch. Using copy-on-write to prevent concurrent modification.
+            currentOrders[marketId] = CopyOnWriteArraySet()
```
imykolenko, line 389: main change. Batch request is now sent from coroutine with debouncing. Concurrent modification occurs when ws order update message arrives at the same time when order cancellations are being prepared for batch.

antonyshchenko (approved):

PR #590, opened 2024-10-29
CHAIN-618: implement active trader reward

bholzman:


backend/src/main/kotlin/xyz/funkybit/apps/api/TestnetChallengeRoutes.kt
```diff
@@ -249,6 +251,15 @@ class TestnetChallengeRoutes {
                 if (!request.principal.user.hasLinkedAccount(UserLinkedAccountType.X)) {
                     cards.add(Card.LinkX)
                 }
+                val tradesByUser = OrderExecutionEntity.countForUser(request.principal.user)
```
imykolenko, line 254: maybe we should add `activeTrader` flag to the user and set it in the response processor, this would be cheaper than counting trades for the active traders

imykolenko (approved):

backend/src/main/kotlin/xyz/funkybit/apps/api/TestnetChallengeRoutes.kt
bholzman: yeah, good point, I'll do that -- will make the check in response processor for active traders faster too

PR #593, opened 2024-10-30
Grant sequencer and garp access to their secrets

antonyshchenko:


imykolenko (approved):
thanks for taking care!

PR #597, opened 2024-10-30
CHAIN-622: use 0_n subscript notation for small prices, add symbol de…

bholzman:
…scription as title to symbol

Screenshot 2024-10-30 at 2 42 57 PM

only applies to prices with leading zeroes after the decimal
Screenshot 2024-10-30 at 2 43 01 PM


bflood366 (approved):
looks great

imykolenko:
Nice 👍

PR #603, opened 2024-10-31
CHAIN-619 Add a ReCaptchaV3 as part of testnet enrollment

imykolenko:
Adds recaptcha v3 to the testnet challenge enrollment flow.

ReCaptcha projects are created in the `chainring.co` google cloud.
localhost - https://www.google.com/recaptcha/admin/site/712069818
testnet - https://www.google.com/recaptcha/admin/site/711959651


integrationtests/build.gradle.kts
```diff
@@ -75,6 +75,7 @@ dependencies {
 
 tasks.test {
     useJUnitPlatform()
+    environment("RECAPTCHA_ENABLED", "false")
```
imykolenko, line 78: recaptcha is disabled during tests execution since we can't obtain token

backend/src/main/kotlin/xyz/funkybit/apps/api/TestnetChallengeRoutes.kt
```diff
@@ -77,38 +82,52 @@ class TestnetChallengeRoutes {
...
+                        )
 
-                    if (DepositEntity.existsForUserAndSymbol(user, symbol)) {
-                        user.testnetChallengeStatus = TestnetChallengeStatus.Disqualified
-                        return@transaction testnetChallengeDisqualifiedError("Disqualified due to prior deposit")
-                    }
+                        ReCaptchaResult.FAILED -> throw RequestProcessingError(
+                            ReasonCode.UnexpectedError,
+                            "Unexpected error. Please try again.",
+                        )
```
antonyshchenko, line 94: The nesting level is becoming too deep. Let's move this into a helper method that would throw `RequestProcessingError`. So here it would then look something like:
```
checkIfHuman(body.token)
```

backend/src/main/kotlin/xyz/funkybit/apps/api/model/TestnetChallenge.kt
```diff
@@ -16,7 +16,10 @@ data class SetNickname(val name: String)
 data class SetAvatarUrl(val url: String)
 
 @Serializable
-data class Enroll(val inviteCode: String?)
+data class Enroll(
+    val inviteCode: String?,
+    val token: String,
```
antonyshchenko, line 21: `token` is a bit too generic, let's maybe rename it to `captchaToken` or something like that

web-ui/src/components/Screens/HomeScreen/testnetchallenge/TestnetChallengeTab.tsx
```diff
@@ -411,6 +410,111 @@ export function TestnetChallengeTab({
...
+        return token
+      } catch (error) {
+        console.error('ReCaptcha execution failed:', error)
+        return null
+      }
+    },
+    [executeRecaptcha]
+  )
+
+  return {
```
antonyshchenko, line 516: Let's move these to some other file if possible

antonyshchenko (approved):
Looks great!

web-ui/src/components/Screens/HomeScreen/testnetchallenge/TestnetChallengeTab.tsx
imykolenko: moved

backend/src/main/kotlin/xyz/funkybit/apps/api/TestnetChallengeRoutes.kt
imykolenko: good point, extracted as `assertHumanUser`

PR #604, opened 2024-10-31
CHAIN-625 ensure user is "logged out" of leaderboard if they disconnect their wallet

imykolenko:
When wallet is disconnected `accountConfigQuery` fails with 401, retries 3 times with 1 second interval, and still keeps previous state.


web-ui/src/components/Screens/HomeScreen.tsx
```diff
@@ -119,12 +119,19 @@ function HomeScreenContent() {
...
   const { isAuthenticated } = useAuth()
 
+  const queryClient = useQueryClient()
   const accountConfigQuery = useQuery({
     queryKey: accountConfigQueryKey,
     queryFn: apiClient.getAccountConfiguration,
     enabled: isAuthenticated
   })
 
+  useEffect(() => {
```
bholzman, line 129: there's this code in TestnetChallengeTab -
```
 // once the wallet is connected, invalidate the accountConfigQuery if it becomes unconnected
  const [walletHasConnected, setWalletHasConnected] = useState(false)
  useEffect(() => {
    if (wallets.connected.length === 0) {
      if (walletHasConnected) {
        setWalletHasConnected(false)
        queryClient.invalidateQueries({ queryKey: accountConfigQueryKey })
      }
    } else if (!walletHasConnected) {
      setWalletHasConnected(true)
    }
  }, [wallets, walletHasConnected, queryClient])
```

I think this can probably go away now, do you think?

bholzman (approved):
Looks good!

web-ui/src/components/Screens/HomeScreen.tsx
imykolenko: you right, thanks! Idea is similar, but reset should be done instead of invalidate because api returns 401.

PR #605, opened 2024-10-31
Make telemetry data accessible in tests

antonyshchenko:
The goal here is to allow checking the number of database calls via tests to make optimization efforts easier to verify and to avoid regressions in the future


integrationtests/src/test/kotlin/xyz/funkybit/integrationtests/api/OrderExecutionTest.kt
```diff
@@ -1045,6 +1047,13 @@ class OrderExecutionTest : OrderBaseTest() {
 
         makerWsClient.close()
         takerWsClient.close()
+
+        val applyOrderBatchTraceIds = getTraceIdsBySpanName("SequencerRequest.Type.ApplyOrderBatch")
+        assertEquals(3, applyOrderBatchTraceIds.size)
+
+        assertDbCallsCountInTrace(30, applyOrderBatchTraceIds[0])
+        assertDbCallsCountInTrace(31, applyOrderBatchTraceIds[1])
+        assertDbCallsCountInTrace(119, applyOrderBatchTraceIds[2])
```
antonyshchenko, line 1056: Here is the interesting part

contracts
bflood366, line 1: looks like your contracts is not updated or something

bflood366 (approved):
looks great

contracts
antonyshchenko: yup, thanks!

PR #606, opened 2024-10-31
CHAIN-560: replenish program balances to account for network fees

bflood366:


backend/src/main/kotlin/xyz/funkybit/core/utils/bitcoin/ArchUtils.kt
```diff
@@ -494,15 +503,17 @@ object ArchUtils {
...
-        rawTx.addOutput(
-            TransactionOutput(
-                params,
-                rawTx,
-                zeroCoinValue,
-                changeAddress.toBitcoinCoreAddress(params),
-            ),
-        )
-        utxos.forEach {
+        (0..<2).forEach { _ ->
```
bflood366, line 506: this change was because I forgot when doing the 2 phase withdrawal to account for the extra in/out for the state transition in the fee estimate - so added it here.

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    override fun runWithLock() {
+        try {
+            val args = nextInvocationArgs
+            transaction {
+                if (args != null) {
+                    checkReplenishment(args.threshold, args.amount)
+                } else {
+                    checkReplenishment(config.replenishmentThreshold, config.replenishmentAmount)
```
antonyshchenko, line 55: Nit: could combine `onfig.replenishmentThreshold` and `config.replenishmentAmount` into the same data class that holds args override and then change `checkReplenishment` to take that data class as param. Then the code here could look like this:

```
checkReplenishment(args ?: config.defaultArgs)
```

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    private fun checkReplenishment(threshold: BigInteger, replenishmentAmount: BigInteger) {
+        //
+        // keeps a running total of withdrawal fees collected - networks fees paid for withdrawals
+        // if the amount is negative and exceeds the threshold, a replenishment tx is sent from
+        // fee payer to the deployed program address to replenish the funds in the program.
+        //
+        val address = DeployedSmartContractEntity.validContracts(ChainId.Arch).firstOrNull()?.proxyAddress ?: return
```
antonyshchenko, line 69: Nit: could store this into a mutable property of the class so that we stop querying the contract address on every run after it becomes available

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
```diff
@@ -0,0 +1,125 @@
...
+        val programDepositAddress = address as BitcoinAddress
+
+        val info = KeyValueStore.getValue("ArchReplenishmentInfo")?.let {
+            Json.decodeFromString<ArchReplenishmentInfo>(it)
```
antonyshchenko, line 73: Let's add a generic method to `KeyValueStore` to store and get data classes

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
bflood366: changed

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
bflood366: done

backend/src/main/kotlin/xyz/funkybit/core/repeater/tasks/ArchBalanceReplenishmentTask.kt
bflood366: yep, thanks good idea.

PR #608, opened 2024-11-01
CHAIN-606 make sure order appears only once on orders changed list in the sequencer response

imykolenko:
Interesting failure mode, order may stuck in 'Partial' state instead of 'Filled'.
Root causes: 
1. order may be present more than once on the `ordersChangedList` in the sequencer response
2. sql batch update does not guarantee ordering, when orderGuid is present several times some updates might be overwritten. Its probabilistic.

How to reproduce: 
 - send a batch of market orders (or limit orders that cross the book) that result into filling an existing limit order IN CHUNKS. several updates on the same order applied in batch, and could result into earlier update being written last.
 - on demo env this requires to put load so that market maker receives order updates with the significant delay and sends new batch w/o cancelling previous, resulting into current batch being matched cross book with previous batch.


sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -393,25 +393,30 @@ object SequencerResponseProcessorService {
...
-                OrderDisposition.AutoReduced -> true
-                else -> null
-            }
-            val newQuantity: BigInteger? = orderChanged.newQuantityOrNull?.toBigInteger()
+        val batchSqlUpdateParams = ordersChanged
+            .groupBy { it.guid }
+            // batch sql update does not guarantee order
+            // ensure we only process the most recent status for each order
+            .map { (_, orders) -> orders.last() }
+            .map { orderChanged ->
```
imykolenko, line 401: safety net, not strictly needed, but if order appears several times on the list we have no ordering guarantee, update might be lost.

sequencer/src/main/kotlin/xyz/funkybit/sequencer/core/Market.kt
```diff
@@ -276,17 +276,20 @@ data class Market(
...
+                    this.guid = execution.counterOrder.guid.value
+                    if (execution.counterOrderExhausted) {
+                        this.disposition = OrderDisposition.Filled
+                    } else {
+                        this.disposition = OrderDisposition.PartiallyFilled
+                        this.newQuantity = execution.counterOrder.quantity.toIntegerValue()
+                    }
+                },
+            )
+        }
```
imykolenko, line 292: actual change, replace counter order changed information if it was matches more than once

imykolenko: yeah, it is not backward compatible

sequencer/src/main/kotlin/xyz/funkybit/sequencer/core/Market.kt
```diff
@@ -276,17 +276,20 @@ data class Market(
...
-                if (execution.counterOrderExhausted) {
-                    this.disposition = OrderDisposition.Filled
-                } else {
-                    this.disposition = OrderDisposition.PartiallyFilled
-                    this.newQuantity = execution.counterOrder.quantity.toIntegerValue()
-                }
-            },
-        )
+        ordersChanged.apply {
+            removeIf { it.guid == execution.counterOrder.guid.value }
```
antonyshchenko, line 280: This would be an `O(N)` operation for every order. How about we just add them without any checks, but remove duplicates in the end? E.g:

```
       return AddOrdersResult(
            ordersChanged.distinctBy { it.guid },
```

bholzman (approved):
Wow, nice find!

antonyshchenko (approved):
Looks great!

sequencer/src/main/kotlin/xyz/funkybit/sequencer/core/Market.kt
imykolenko: since we need last version it should be `ordersChanged.reversed().distinctBy { it.guid }.reversed()`, not sure it will be more efficient.
I will leave it as is for now.

PR #609, opened 2024-11-01
CHAIN-630 avoid using KV store for order book snapshot sequence number

imykolenko:
sequence number is stored directly with order book snapshot instead of KV store, removes 2 db calls.


antonyshchenko (approved):
Nice! 👍

PR #610, opened 2024-11-01
Fix Sequencer checkpointing

antonyshchenko:
In demo env Sequencer was failing on response validation of first message after the checkpoint. Turned out that we include the first message of new cycle into the checkpoint and when we restore from checkpoint we try to process this message again, but having state that already includes the effect of this message. 

This was happening because of the wrong assumption regarding the moment at which input tailer moves to the next cycle. Previously we were checking if cycle has changed and were making a checkpoint, but this would happen after we have processed the first message of the cycle. The correct way would be to read the first message, detect that cycle has changed, make checkpoint and then process this message.

Tests were not detecting this because SequencerApp was running without `strictReplayValidation`


imykolenko (approved):
👍

bholzman (approved):
Nice!

PR #613, opened 2024-11-01
CHAIN-608: mempool space client retries

bflood366:


backend/src/main/kotlin/xyz/funkybit/apps/ring/BitcoinDepositHandler.kt
```diff
@@ -67,7 +67,12 @@ class BitcoinDepositHandler(
 
     private fun refreshPendingDeposit(pendingDeposit: DepositEntity, currentBlockHeight: Long) {
         logger.info { "Refreshing pending deposit ${pendingDeposit.guid}, currentBlockHeight=$currentBlockHeight" }
-        val tx = mempoolSpaceClient.getTransaction(pendingDeposit.transactionHash)
+        val tx = try {
+            mempoolSpaceClient.getTransaction(pendingDeposit.transactionHash)
+        } catch (e: Exception) {
+            logger.warn { "failed to get transaction $e" }
```
bflood366, line 73: each run of the deposit handler is within a single transaction, so if there was like 20 deposits pending and the prior 19 were processed ok, but the api call failed for the 20th one, all of them would have been rolled back. Now if the API call fails on the 20th one it just skips it, but any prior ones processed would be updated. Same logic was added to withdraw handler

bholzman (approved):
Looks good

PR #614, opened 2024-11-01
send telemetry to honeycomb

imykolenko:


backend/src/main/kotlin/xyz/funkybit/Main.kt
```diff
@@ -5,15 +5,15 @@ import xyz.funkybit.apps.api.ApiApp
...
+import xyz.funkybit.core.telemetry.telemetry
 import kotlin.system.exitProcess
 
 private val logger = KotlinLogging.logger {}
 
 fun main(args: Array<String>) {
     logger.info { "Starting with args: ${args.joinToString(" ")}" }
 
-    val telemetry = sentryTelemetry
+    val telemetry = telemetry
```
antonyshchenko, line 16: Maybe just get rid of the `telemetry` val here

src/main/kotlin/xyz/funkybit/LocalMain.kt
```diff
@@ -17,7 +17,7 @@ import kotlin.system.exitProcess
 
 fun main(args: Array<String>) {
     val logger = KotlinLogging.logger {}
-    val telemetry = sentryTelemetry
+    val telemetry = telemetry
```
antonyshchenko, line 20: And here

antonyshchenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/Main.kt
imykolenko: sure, makes sense, after renaming local var is obsolete.

PR #617, opened 2024-11-01
CHAIN-631: use keys from secrets manager for arch program/withdraw accounts

bflood366:
- also made the arch deployment process idempotent 
- added decryption key to testnet


backend/src/main/kotlin/xyz/funkybit/core/model/EncryptedString.kt
```diff
@@ -14,11 +13,11 @@ import java.security.interfaces.ECPrivateKey
 import java.security.spec.PKCS8EncodedKeySpec
 
 private val bcProvider = BouncyCastleProvider()
-private val dataEncryptionKey = ECPublicKeyDecoder.fromHexEncodedString(System.getenv("DATA_ENCRYPTION_KEY") ?: "0x0443975c5a5cf6caaf3cc846dac466f1e81321f5dc36f97fc2528c68449381a828520946caf970fa3f7e508a56d902c2f4405aed01972f5e3bb382f84bff56b042")
 private val dataDecryptionKey = run {
     val privateKeyBytes = (System.getenv("DATA_DECRYPTION_KEY") ?: "0x308193020100301306072a8648ce3d020106082a8648ce3d0301070479307702010104206d251224ebaeedfd87cca2093ed45c9d3516f7bb8d9a66567378455524944d58a00a06082a8648ce3d030107a1440342000443975c5a5cf6caaf3cc846dac466f1e81321f5dc36f97fc2528c68449381a828520946caf970fa3f7e508a56d902c2f4405aed01972f5e3bb382f84bff56b042").toHexBytes()
     KeyFactory.getInstance("EC", bcProvider).generatePrivate(PKCS8EncodedKeySpec(privateKeyBytes)) as ECPrivateKey
 }
+private val dataEncryptionKey = ECIESManager.derivePublicKeyFromPrivateKey(dataDecryptionKey)
```
bflood366, line 20: I change this to derive the encryption key from the decryption key so we didn't have to store 2 different values per env.

terraform/modules/ecs_task/main.tf
```diff
@@ -52,7 +52,8 @@ resource "aws_iam_role_policy" "ecs_execution_role_policy" {
         ],
         Resource = [
           "arn:aws:secretsmanager:${var.aws_region}:${local.account_id}:secret:${var.name_prefix}/${var.task_name}/*",
-          "arn:aws:secretsmanager:${var.aws_region}:${local.account_id}:secret:slack-error-reporter-token-*"
+          "arn:aws:secretsmanager:${var.aws_region}:${local.account_id}:secret:slack-error-reporter-token-*",
+          "arn:aws:secretsmanager:${var.aws_region}:${local.account_id}:secret:${var.name_prefix}/data-decryption-key-*"
```
bholzman, line 56: may be better to create a `${var.name_prefix}/shared` secret path and put the data-decryption-key there. that way we could also put things like the HONEYCOMB_API_KEY where we right now have multiple copies of in secrets manager.

bholzman (approved):
Looks good!

terraform/modules/ecs_task/main.tf
bflood366: ok, change this `${var.name_prefix}/shared/*` and placed `data-decryption-key` under this path in secrets mgr

PR #620, opened 2024-11-04
Cache symbols and markets in SequencerResponseProcessor

antonyshchenko:


imykolenko (approved):
Looks great!

honeycomb is activated on demo, we should be able to see processing time reduction immediately!

PR #633, opened 2024-11-05
Get market chain ids without a DB call

antonyshchenko:


backend/src/main/resources/exchangeprogram.so
antonyshchenko, line 1: Result of running `make contracts`

PR #636, opened 2024-11-05
CHAIN-616: enable arch for testnet

bflood366:


backend/src/main/kotlin/xyz/funkybit/core/blockchain/bitcoin/ArchNetworkClient.kt
```diff
@@ -15,7 +15,9 @@ class ArchNetworkClient(telemetry: OpenTelemetry) : JsonRpcClientBase(
     System.getenv("ARCH_NETWORK_RPC_URL") ?: "http://localhost:9002",
     KotlinLogging.logger {},
     telemetry,
-    null,
+    System.getenv("ARCH_API_KEY")?.let {
```
bflood366, line 18: I initially added this since maestro, the arch RPC provider requires an api key - but we are just using RPC node directly from Arch for now - so this is not used for now but will leave it here.

backend/src/main/kotlin/xyz/funkybit/core/blockchain/bitcoin/JsonRpcClientBase.kt
```diff
@@ -75,30 +84,36 @@ open class JsonRpcClientBase(val url: String, val logger: KLogger, telemetry: Op
...
-                }?.let {
-                    throw JsonRpcException(it)
+                    throw Exception("Got an error (${httpResponse.code}) - $httpResponseBody")
                 }
-                throw Exception("Got an error (${httpResponse.code}) - $httpResponseBody")
             }
+        } catch (e: Exception) {
+            throw e
+        } finally {
+            httpResponse.body?.close()
```
bflood366, line 116: I added this since I noticed these warnings in logs and this seemed to fix it 
```
WARNING: A connection to https://mempool.space/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```

antonyshchenko (approved):
👍

PR #637, opened 2024-11-06
Chain 647/limits

imykolenko:
- update limits in 1 db call, reuse calculation result for ws
 - ws limits message includes only changed limits


integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/Utils.kt
```diff
@@ -110,7 +112,12 @@ fun assertFee(expectedAmount: AssetAmount, amount: BigInteger, symbol: Symbol) {
 }
 
 fun verifyApiReturnsSameLimits(apiClient: TestApiClient, wsMessage: Limits) {
-    assertEquals(wsMessage.limits, apiClient.getLimits().limits)
+    val apiLimits = apiClient.getLimits().limits
+    val wsLimits = wsMessage.limits
+    assertTrue { wsLimits.isNotEmpty() }
+    wsLimits.forEach { marketLimits: MarketLimits ->
+        assertEquals(apiLimits.find { it.marketId == marketLimits.marketId }, marketLimits)
```
antonyshchenko, line 119: Function name became misleading. Please consider renaming to something like `verifyApiContainsSameLimits`

sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -271,13 +272,18 @@ object SequencerResponseProcessorService {
 
         val broadcasterNotifications = mutableListOf<BroadcasterNotification>()
 
+        val userWalletsMap = UserEntity.getWithWalletsBySequencerAccountIds(
```
antonyshchenko, line 275: 👍

sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/services/SequencerResponseProcessorService.kt
```diff
@@ -644,39 +649,38 @@ object SequencerResponseProcessorService {
...
-        usersToNotifyAboutLimitsChanges.forEach { wallet ->
-            broadcasterNotifications.add(
-                BroadcasterNotification.limits(wallet),
+        val userLimitsUpdates: List<Pair<UserId, MarketLimits>> = limitsUpdates.map {
+            val marketId = MarketId(it.marketId)
+            val (user, userWallets) = userWalletsMap[it.account]!!
+
+            getMarket(marketId).chainIds()
+                .mapNotNull { chainId -> userWallets.find { wallet -> wallet.isForChain(chainId) } }
+                .forEach { wallet -> usersToNotifyAboutLimitsChanges.add(wallet.userGuid.value) }
```
antonyshchenko, line 666: Nit: not sure about populating this inside the `map` function. We could calculate it from `userLimitsUpdates` insted, e.g:
```
val usersToNotifyAboutLimitsChanges = userLimitsUpdates.map { it.first }.toSet()
```

Alternatively we could make `userLimitsUpdates` a mutable list and change `map` to `forEach`. The second option might be marginally faster.

antonyshchenko (approved):
Looks great!

sequencer/src/main/kotlin/xyz/funkybit/sequencer/apps/services/SequencerResponseProcessorService.kt
imykolenko: good catch, I missed it. Since users are notified with the calculated limit values `usersToNotifyAboutLimitsChanges` is no longer in used. I just drop it.

PR #640, opened 2024-11-06
CHAIN-648: add a "connect" button to connected wallet drop-down if a …

bholzman:
…wallet type is not connected


bflood366 (approved):
Looks good

PR #645, opened 2024-11-07
testnet challenge test fix

bholzman:


imykolenko (approved):
Balance monitor test failed. It seems the minimum enrollment balance check is not passing due to the increased gas cost

PR #647, opened 2024-11-07
Avoid unnecessary orders fetching in SequencerResponseProcessor

antonyshchenko:
To send out notifications we were fetching orders from DB after updating them with batch update, but that is unnecessary since we have all the information we need already in memory.


backend/src/main/kotlin/xyz/funkybit/core/model/db/Order.kt
```diff
@@ -260,10 +245,24 @@ class OrderEntity(guid: EntityID<OrderId>) : GUIDEntity<OrderId>(guid) {
...
+        fun batchUpdate(assignments: List<UpdateOrderAssignment>, timestamp: Instant) {
+            TransactionManager.current().exec(
+                """
+                UPDATE "${OrderTable.tableName}" AS t
+                SET 
+                    ${OrderTable.status.name}         = COALESCE(v.status::OrderStatus, t.${OrderTable.status.name}),
+                    ${OrderTable.autoReduced.name}    = COALESCE(v.auto_reduced::boolean, t.${OrderTable.autoReduced.name}),
+                    ${OrderTable.amount.name}         = COALESCE(v.amount::numeric, t.${OrderTable.amount.name}),
+                    ${OrderTable.originalAmount.name} = COALESCE(v.original_amount::numeric, t.${OrderTable.originalAmount.name}),
+                    ${OrderTable.updatedAt.name}      = '$timestamp'
```
imykolenko, line 257: since sequencer response processor was updated to re-use already loaded OrderEntities it looks like this sql does not need to handle `null` values anymore.
```
SET 
                    ${OrderTable.status.name}                = v.status::OrderStatus,
                    ${OrderTable.autoReduced.name}    = v.auto_reduced::boolean, 
                    ${OrderTable.amount.name}              = v.amount::numeric, 
                    ${OrderTable.originalAmount.name} = v.original_amount::numeric, 
                    ${OrderTable.updatedAt.name}         = '$timestamp'
```

imykolenko (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/Order.kt
imykolenko: hm, and then it can be generalized to instrumentedBatchUpdate, wdyt?

PR #650, opened 2024-11-08
Update exposed

imykolenko:


integrationtests/build.gradle.kts
```diff
@@ -61,12 +61,12 @@ dependencies {
     implementation("io.grpc:grpc-netty:$grpcProtobufVersion")
     implementation("net.openhft:chronicle-queue:5.22.28")
```
imykolenko, line 62: `5.22.28` version is from 2022. Newer versions have `ea` (net.openhft:chronicle-queue:5.26ea9) which suggests `early access` probably for evaluation reason. Where can we obtain newer version?

https://central.sonatype.com/artifact/net.openhft/chronicle-queue/versions

backend/build.gradle.kts
```diff
@@ -1,10 +1,9 @@
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    kotlin("jvm") version "1.9.23"
-    kotlin("plugin.serialization") version "1.9.23"
+    kotlin("jvm") version "2.0.21"
```
antonyshchenko, line 3: 🔥

backend/src/main/kotlin/xyz/funkybit/core/model/db/ArchAccount.kt
```diff
@@ -82,9 +83,18 @@ sealed class AccountSetupState {
 object ArchAccountTable : GUIDTable<ArchAccountId>("arch_account", ::ArchAccountId) {
     val createdAt = timestamp("created_at")
     val updatedAt = timestamp("updated_at").nullable()
-    val utxoId = varchar("utxo_id", 10485760).uniqueIndex()
-    val creationTxId = varchar("creation_tx_id", 10485760).nullable()
-    val encryptedPrivateKey = varchar("encrypted_private_key", 10485760)
+    val utxoId = varchar("utxo_id", 10485760).uniqueIndex().transform(
+        wrap = { BitcoinUtxoId(it) },
```
antonyshchenko, line 87: Nice! Is this a new feature of Exposed?

backend/src/main/kotlin/xyz/funkybit/core/model/db/EntityId.kt
```diff
@@ -1,20 +1,100 @@
...
+// class EntityIdColumnType<T : EntityId>(val deserialize: (String) -> T) : VarCharColumnType(colLength = 10485760) {
+//    override fun valueFromDB(value: Any): Any {
+//        return when (val v = super.valueFromDB(value)) {
+//            is String -> deserialize(v)
+//            else -> v
+//        }
+//    }
+//
+//    override fun notNullValueToDB(value: Any): Any = value.toString()
+// }
```
antonyshchenko, line 100: Let's remove this

antonyshchenko (approved):
Great job! 👍

backend/src/main/kotlin/xyz/funkybit/core/model/db/ArchAccount.kt
imykolenko: yes, one of main changes, transformation is moved from entity to DSL. This allows to use value classes directly in queries. 

Types got stricter, but IDs are no longer comparable, needs to be unwrapped.

Upsert API got better, even promises to read value returned from the DB 😜

PR #651, opened 2024-11-08
CHAIN-661 add bitcoin deposits to mocker

bflood366:


mocker/src/main/kotlin/xyz/funkybit/mocker/MockerAppMain.kt
```diff
@@ -146,6 +146,7 @@ class MockerApp(
                         baseAssetAmount = params.initialBaseBalance * BigDecimal(100),
                         quoteAssetAmount = params.initialBaseBalance * params.priceBaseline * BigDecimal(100),
                         keyPair = WalletKeyPair.EVM.fromPrivateKeyHex(params.makerPrivateKeyHex),
+                        bitcoinKeyPair = WalletKeyPair.Bitcoin.fromPrivateKeyHex(params.makerPrivateKeyHex),
```
bholzman, line 149: I was envisioning we'd have a separate private key for the bitcoin wallet, but I suppose this is fine and easier to manage

mocker/src/main/kotlin/xyz/funkybit/mocker/core/Actor.kt
```diff
@@ -155,19 +159,45 @@ abstract class Actor(
                                 } else null
                             } else null
                         }
+                    } + config.bitcoinChain.let { bitcoinChain ->
```
bholzman, line 162: we don't have the bitcoin chain in demo, so this should be optional, no?

mocker/src/main/kotlin/xyz/funkybit/mocker/core/Actor.kt
```diff
@@ -155,19 +159,45 @@ abstract class Actor(
...
+                        logger.debug { "Starting authorization for bitcoin wallet" }
+                        // setup bitcoin wallet and authorize here
```
bholzman, line 164: seems awkward to have the authorization here ... would it be better to be part of the `apiClient` setup in some way?

Makefile
```diff
@@ -79,3 +79,5 @@ run_telegrambot:
 run_ui:
 	cd web-ui && pnpm install && pnpm run dev
 
+run_mocker_btc_bitcoin:
+	FAUCET_POSSIBLE=0 MARKETS=BTC:1337/BTC:bitcoin BTC_1337_BTC_bitcoin_PRICE_BASELINE=1.01 BTC_1337_BTC_bitcoin_MAKER_PRIVATE_KEY_HEX=0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6 BTC_1337_BTC_bitcoin_TAKERS=0 API_URL=http://localhost:9000 ./gradlew :mocker:run
```
bholzman, line 83: do we need this private key checked in?

bholzman (approved):
Looks good

PR #655, opened 2024-11-08
CHAIN-671 fix for allowing multiple deposits/withdrawals for same wal…

bflood366:
…let/symbol to be in same batch


bholzman (approved):
Looks good!

bholzman (approved):
Looks good x2!

PR #660, opened 2024-11-11
Better error message on wallet authorization

antonyshchenko:
In case of trying to authorize a wallet that was already authorized by another wallet the user would see "Authorized wallet address is already in use" error coming from the backend


bflood366 (approved):
looks good

antonyshchenko:
Changed to show by whom the wallet is already being used:

Screenshot 2024-11-11 at 17 33 49

PR #663, opened 2024-11-11
Calculate testnet challenge deposit limits correctly when user has used 'withdraw all feature'

antonyshchenko:


backend/src/main/kotlin/xyz/funkybit/core/utils/TestnetChallengeUtils.kt
```diff
@@ -38,7 +41,17 @@ object TestnetChallengeUtils {
 
         val depositSymbol = symbols.first { it.name == depositSymbolName }
 
-        val withdrawalAmountsSumColumn = WithdrawalTable.amount.sum()
+        val withdrawalAmountsSumColumn = Case()
```
bholzman, line 44: Do we always populate actual amount? Would it be better to always use it?

bholzman (approved):
Looks good!

backend/src/main/kotlin/xyz/funkybit/core/utils/TestnetChallengeUtils.kt
antonyshchenko: Yeah, we do. Changed as you suggest

PR #664, opened 2024-11-11
CHAIN-682: retrieve all price feed prices in same process with one call

bholzman:


backend/src/main/kotlin/xyz/funkybit/core/utils/PriceFeed.kt
```diff
@@ -5,75 +5,100 @@ import kotlinx.serialization.json.Json
...
+                                    Json.decodeFromString<Map<String, Map<String, Double>>>(response.body!!.string())
+                                prices.keys.forEach { marketId ->
+                                    rawBasesAndQuotesByMarket[marketId]?.let { (marketBase, marketQuote) ->
+                                        receivedPrices[marketBase]?.let { basePrices ->
+                                            basePrices[marketQuote]?.let { prices[marketId] = it }
+                                        }
+                                    }
+                                }
+                                val immutablePrices = prices.toMap()
+                                callbacksByPriceFeed.values.forEach { it(immutablePrices) }
```
bflood366, line 79: if I grok how this is working (and I might be missing something), then each individual who registers a callback for a specific market will get updates for all the markets that are being monitored by the thread started by the companion. I guess then it depends how the callers of this handle the prices in the callback since I think they would get prices for markets they don't expect.

bholzman: no, you're right -- I thought the mock maker would just ignore prices it didn't care about, but I missed that it will actually throw an exception if it sees a price it wasn't expecting. I'll change how the price feed works to only deliver the prices each consumer is interested in

bflood366 (approved):
looks great

PR #666, opened 2024-11-12
Add .kotlin folder to .gitignore

antonyshchenko:
Seems like this folder is supposed to be in `.gitignore`:

https://youtrack.jetbrains.com/issue/KTIJ-30076/Add-.kotlin-folder-in-.gitignore-in-the-wizard-generated-Gradle-projects


imykolenko (approved):
nice!

PR #673, opened 2024-11-13
Migrate from WalletConnect to Reown AppKit

antonyshchenko:


web-ui/src/utils/addSymbolsToWallet.ts
```diff
@@ -1,5 +1,13 @@
 import { apiClient } from 'apiClient'
 
+import { MetaMaskInpageProvider } from '@metamask/providers'
+
+declare global {
+  interface Window {
+    ethereum?: MetaMaskInpageProvider
+  }
+}
+
```
antonyshchenko, line 10: Without this type-checking was breaking on `window.ethereum?.request` call bellow

bholzman (approved):
Looks good!

PR #678, opened 2024-11-13
CHAIN-686 check balance before deposit using noFaucetNativeDepositRatio

imykolenko:


mocker/src/main/kotlin/xyz/funkybit/mocker/core/Actor.kt
```diff
@@ -165,18 +170,30 @@ abstract class Actor(
...
+
                                 if (balanceToDeposit > BigInteger.ZERO) {
                                     bitcoinWallet.depositNative(balanceToDeposit).also {
-                                        logger.debug { "deposit response is ${it.deposit}" }
+                                        logger.debug { "$id: Deposit response is ${it.deposit}" }
+                                    }
+                                    val depositToDepositedRatio = balanceToDeposit.toBigDecimal().divide(desiredDepositedBalance.toBigDecimal())
+                                    if (depositToDepositedRatio > noFaucetNativeDepositSkipConfirmationRatio || depositedBalance == BigInteger.ZERO) {
+                                        deposits += 1
+                                    } else {
```
imykolenko, line 193: I thought it might be worse to skip awaiting for confirmation when previously deposited and now filling back less than 5%. This may apply in case of mocker restart and balance slightly changes due to price move.

integrationtests/src/main/kotlin/xyz/funkybit/integrationtests/utils/Wallet.kt
```diff
@@ -164,9 +164,9 @@ class Wallet(
         waitForTransactionReceipt(sendDepositTx(assetAmount))
 
     fun sendDepositTx(assetAmount: AssetAmount): TxHash = if (assetAmount.symbol.contractAddress == null) {
-        sendNativeDepositTx(assetAmount.amount.toFundamentalUnits(assetAmount.symbol.decimals))
+        sendNativeDepositTx(assetAmount.inFundamentalUnits)
```
bholzman, line 167: 👍

mocker/src/main/kotlin/xyz/funkybit/mocker/core/Actor.kt
bholzman: yes, I definitely think it makes sense to not wait for confirmation if we only have to deposit a little

bholzman (approved):
Looks great!

PR #698, opened 2024-11-19
CHAIN-712: implement alerts

bholzman:
Screenshot 2024-11-18 at 11 40 53 PM


web-ui/src/utils/alertUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+    'funkybit-testnet-dismissed-alerts'
+  )
+  let alertObj
+  if (alerts === null || alerts === 'null') {
+    alertObj = {}
+  } else {
+    alertObj = JSON.parse(alerts)
+  }
+  alertObj[id] = true
+  window.localStorage.setItem(
```
antonyshchenko, line 24: This is fine as a quick solution, but user might get confused if they open the app in another browser.

antonyshchenko (approved):

web-ui/src/utils/alertUtils.ts
bholzman: I mean, they'll get a few alerts they already got, but doesn't seem like a big deal

PR #701, opened 2024-11-19
CHAIN-715: add lock state to bitcoin withdrawal+deposit mission

bholzman:
Screenshot 2024-11-19 at 12 12 55 PM
Screenshot 2024-11-19 at 12 12 58 PM


antonyshchenko (approved):
Looks good!

PR #703, opened 2024-11-19
CHAIN-714: add points leaderboard

bholzman:
Screenshot 2024-11-19 at 2 00 30 PM
Screenshot 2024-11-19 at 2 00 42 PM


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -44,6 +45,7 @@ enum class TestnetChallengePNLType {
     DailyPNL,
     WeeklyPNL,
     OverallPNL,
+    FunkyBits,
```
antonyshchenko, line 48: Not sure about having this here since it kind of introduces a special case in terms of whether we create rows in `testnet_challenge_pnl` table with this type. Could we perhaps have a different enum named `LeaderboardType` in the API model which would include this case?

web-ui/src/components/Screens/HomeScreen/testnetchallenge/Leaderboard.tsx
```diff
@@ -208,8 +202,14 @@ function Board() {
...
-                <td className="w-1/3 py-4 text-sm font-bold">Balance</td>
-                <td className="py-4 pr-4 text-sm font-bold">PNL</td>
+                <td className="w-1/3 py-4 text-sm font-bold">
+                  {selectedLeaderboard === 'FunkyBits'
+                    ? 'Funky Bits'
+                    : 'Balance'}
+                </td>
+                {selectedLeaderboard !== 'FunkyBits' && (
+                  <td className="py-4 pr-4 text-sm font-bold">PNL</td>
+                )}
```
antonyshchenko, line 212: Here it seems like two different types are combined into a single struct (PNL leaderboard entry and FB leaderboard entry). I'm generally not a fan of modelling things this way, but since we are in a hurry to push this out ASAP let's keep it as is.

bholzman: yeah, you're right, this could be modeled more cleanly, I'll make an update

PR #704, opened 2024-11-19
Always show user's place in the leaderboard

antonyshchenko:


backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -139,7 +173,7 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
                     -- initial testnet challenge symbol deposit per user
                     testnet_challenge_initial_deposits AS (SELECT deposit_guid
                                                                 FROM (SELECT d.${DepositTable.guid.name} AS deposit_guid, 
-                                                                             ROW_NUMBER() OVER (PARTITION BY w.${WalletTable.userGuid.name} ORDER BY d.${DepositTable.createdAt.name} ASC) AS rn
+                                                                             DENSE_RANK() OVER (PARTITION BY w.${WalletTable.userGuid.name} ORDER BY d.${DepositTable.createdAt.name} ASC) AS rn
```
antonyshchenko, line 176: I believe `DENSE_RANK()` is better here as it would allow entries with same score to have the same rank

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
```diff
@@ -126,6 +132,34 @@ class TestnetChallengePNLEntity(guid: EntityID<TestnetChallengePNLId>) : GUIDEnt
...
         }
 
+        fun updateRanks() {
+            TestnetChallengePNLType.entries.forEach { pnlType ->
+                TransactionManager.current().exec(
+                    """
+                    WITH ranked_users AS (
+                      SELECT 
+                        pnl.${TestnetChallengePNLTable.userGuid.name} AS user_guid,
+                        pnl.${TestnetChallengePNLTable.type.name} AS type,
```
bholzman, line 142: do we need to select type here?

bholzman (approved):
Looks great!

backend/src/main/kotlin/xyz/funkybit/core/model/db/TestnetChallengePNL.kt
antonyshchenko: good point, we don't

